PR #3, opened 2020-07-25
Add support for composition

patriksvensson:
Adds API for composing elements programmatically.


patriksvensson:
Going to rethink this a bit and submit a new implementation.

PR #25, opened 2020-08-07
Update regex to correctly identify Windows 10

AdmiringWorm:
When running .NET Core 2.1 or .NET Framework 4.6.1, the used regex for
detecting Windows 10 TrueColor support does not match the OSDescription.
The reason for this is because on those frameworks the description has
trailing whitespaces, but they do not on .NET Core 3.1.
This pull request updates the regex to ignore any trailing whitespaces.


patriksvensson:

PR #26, opened 2020-08-07
Correct check for color being the default color or not

AdmiringWorm:
This pull reqest updates the Color class to also take into account of wether one of the colors that are being compared is the default color or not.

Additionally, I added examples for background color in the Samples console project


patriksvensson:
@AdmiringWorm Thanks for this! Appreciated! 👍

PR #76, opened 2020-09-10
Add link support for supported terminals

### LINKED ISSUE #75: Support ANSI Hyperlinks in output
It would be handy if there were support in spectre.console to output ANSI hyperlinks (clickable links) in the terminals that support this feature.

While there are currently only a handful of terminals that currently support this.
Current state of support (per 2020-05-31) can be seen here: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

From my own testing, I have found that GNOME Terminal supports these links, and partially Windows Terminal (mainly only that it won't choke when ANSI hyperlinks are used).

patriksvensson:
Also refactors the code quite a bit, to make it a bit more
easier to add features like this in the future.

Closes #75


patriksvensson:
@AdmiringWorm Can you take this PR for a spin in a terminal which supports links?
There's a new example called `Links` under `./examples/Links`. 

Could you also grab a screenshot of it by any chance? Thanks!

AdmiringWorm:
I took it for a spin. It seems the link isn't going through:

Here is why I got:
image

And as an example, here is what is expected for the ANSI Links
image

The tooltip happening in the screenshot is caused by hovering over the link (and there is a dotted line when not hovering over the link).
But these can be terminal implementation specifics, I think.

patriksvensson:
@AdmiringWorm Think I might have found something. Could you pull the latest and try again?

AdmiringWorm:
@patriksvensson just did, looks to be working as it should:

image

patriksvensson:
Could you try adding a color like `yellow` to the link markup and see if it still looks like it should?

Sorry, will get my linux box up and running again asap so I won't bother you more! 😅

AdmiringWorm:
@patriksvensson sure thing:

<details>
<summary>using [yellow link]:</summary>

image

</details>

<details>
<summary>Using [yellow][link]</summary>

image

</details>

<details>
<summary>Using [link][yellow]</summary>

image

</details>

Aything else you want me to test?

btw, I noticed that using `[link]https://patriksvensson.se][/]!` didn't work, do you plan to add that as well (not a problem if not).

patriksvensson:
@AdmiringWorm Perfect. Seems like it's working now. I haven't thought about that scenario but probably something we want to support. Adding it!

PR #79, opened 2020-09-15
Add emoji code support

khellang:
Was just playing around. It's yours if you want it 😄

I have no clue where to plug it in, so I just did the replacement in `Segment` and added some emoji to the `Info` example 😅

I'll try to write some tests for the emoji replacement soon.

This is currently what the info sample looks like in my Rider terminal:

image


patriksvensson:
@khellang Ooooh, awesome! I very much want this 😄 

I'm not sure where the correct place is to add it as well, but I would say `Segment` is OK by me.
Or perhaps it should be part of `Markup` parsing? Anyway, for now, putting it in `Segment` will do fine.

Thanks!

patriksvensson (approved):

patriksvensson:

khellang:
Awesome. Perfect opportunity to test the new wcwidth implementation 😉

patriksvensson:
@khellang Exactly! 😄

PR #87, opened 2020-10-01
Update Docs Readme

khalidabuhakmeh:
Added some more context to the docs readme for
folks looking to contribute. Also added the license
so its clear what the license of the docs site is. As
a small tweak I removed a duplicate string value and
use the Constant.


patriksvensson (approved):

patriksvensson:
Awesome @khalidabuhakmeh! There can never be enough documentation 😄

Issue #93, opened 2020-10-06
Grids/Table: Exception gets thrown when console width is very small


toburger:
An Exception gets thrown when running the Grids or Table example on a very small console window:

Grid:
> System.InvalidOperationException: Text folding failed since 'take' was zero.

Table:
> System.ArgumentOutOfRangeException: Length cannot be less than zero. (Parameter 'length')

![image](https://user-images.githubusercontent.com/728782/95180118-d515f280-07c1-11eb-91e4-b2d57e1b5f75.png)

I don't expect pretty output but at least the program shouldn't crash.

patriksvensson:
@toburger Yup, it shouldn't crash. Will take a look at it when I have some free time.

patriksvensson:
@toburger @khalidabuhakmeh This should be fixed with #99.
I've just pushed a new release, so it will be available on NuGet in about 15 minutes.

toburger:
Thx! Panel is now working, but Tables throws still an exception:

```
Unhandled exception. System.ArgumentOutOfRangeException: Length cannot be less than zero. (Parameter 'length')
   at System.String.Substring(Int32 startIndex, Int32 length)
   at Spectre.Console.Rendering.Segment.SplitOverflow(Segment segment, Nullable`1 overflow, RenderContext context, Int32 width) in C:\Workspace\oss\spectre.console\src\Spectre.Console\Rendering\Segment.cs:line 312
   at Spectre.Console.Paragraph.SplitLines(RenderContext context, Int32 maxWidth) in C:\Workspace\oss\spectre.console\src\Spectre.Console\Widgets\Paragraph.cs:line 245
   at Spectre.Console.Paragraph.Render(RenderContext context, Int32 maxWidth) in C:\Workspace\oss\spectre.console\src\Spectre.Console\Widgets\Paragraph.cs:line 141
   at Spectre.Console.Rendering.Renderable.Spectre.Console.Rendering.IRenderable.Render(RenderContext context, Int32 maxWidth) in C:\Workspace\oss\spectre.console\src\Spectre.Console\Rendering\Renderable.cs:line 19
   at Spectre.Console.Markup.Render(RenderContext context, Int32 maxWidth) in C:\Workspace\oss\spectre.console\src\Spectre.Console\Widgets\Markup.cs:line 47
[...]

PR #99, opened 2020-10-08
Don't throw when console is small

khalidabuhakmeh:
this just returns an empty collection when
take is 0. It leads to some strange output, but
it doesn't blow up.

 #93

image


patriksvensson:
@khalidabuhakmeh Thanks! This will do fine until addressed properly. Much appreciated!

PR #100, opened 2020-10-11
Add table overflow splitting

patriksvensson:
This PR adds support for table overflow splitting.
Splitting strategies are opt-in and listed below.

## Default

Do nothing to prevent bad rendering.

- [x] Implemented
    
image

## Drop

Start dropping columns until content fit.

- [x] Implemented

image

## Ellipsis

Display columns that won't fit with an ellipsis.

- [ ] Implemented

image

## Column splitting

Split columns in half until all columns fit.

- [ ] Implemented

image

## Row splitting

Split rows so all columns are grouped together even if not all fit.
Introduce row separators for this to make it less confusing?

- [ ] Implemented

image


khalidabuhakmeh:
All these are great options. It would be awesome to set up a global config in one place to adjust the output based on environment.

I imagine on CI I would want default behavior.

PR #107, opened 2020-10-17
I want to add the documentation for Japanese edition

yamatoya:
This library is very useful. If there was Japanese documentation, the library would be growth.

### DIFF
--- a/docs/input/_layout.cshtml
+++ b/docs/input/_layout.cshtml
@@ -1,209 +1,279 @@
 <!DOCTYPE html>
 <html>
-    <head>
-        <meta charset="utf-8">
-        <meta http-equiv="X-UA-Compatible" content="IE=Edge">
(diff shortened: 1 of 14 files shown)


patriksvensson:
@yamatoya I'm delighted with this pull request, and I really appreciate the time you put into this, but it comes with some problems for me. 

I don't know Japanese, so any changes to the documentation will be out of date in the Japanese version, and the documentation will change a lot once Spectre.Console stabilizes a bit.

Perhaps another approach (for the time being) would be to add a localized README as Rich has for Chinese (See https://github.com/willmcgugan/rich/blob/master/README.cn.md) with some additional information on how to get started for Japanese users? We could link to this `README.jp.md` from `README.md` (see image below)?

image

yamatoya:
Thank you for your suggestion. I also agree it and I create another PR.

PR #153, opened 2020-11-29
Interlaced Canvas Graphics

rogeralsing:
Proof of Concept PR for interlaced graphics.
The idea is that Canvas has a `CanvasRenderMode` which decides if it renders using a full ' ' space block, or if it uses a '▄' char with fore and back-color to get two pixels in one char.

No idea if this is something that is of value for the project, but it was fun to play with :-)

**Before:**
Skärmavbild 2020-11-29 kl  18 46 05

**After:**
Skärmavbild 2020-11-29 kl  18 58 43

### DIFF
--- a/examples/Canvas/Program.cs
+++ b/examples/Canvas/Program.cs
@@ -24,12 +24,21 @@ namespace CanvasExample
             });
 
             image.MaxWidth(32);
-            Render(image, "Image from file (16 wide)");
+            Render(image, "Image from file (32 wide)");
 
-            // Draw image again, but without render width
-            image.NoMaxWidth();
-            image.Mutate(ctx => ctx.Grayscale().Rotate(-45).EntropyCrop());
-            Render(image, "Image from file (fit, greyscale, rotated)");
+
+            var image2 = new CanvasImage("cake.png", CanvasRenderMode.Interlaced);
+
+            // Draw image without render width
+            image2.NoMaxWidth();
+            image2.BilinearResampler();
+            image2.Mutate(ctx => ctx.Grayscale().Rotate(-45).EntropyCrop());
+            image2.Mutate(c =>
+            {
+                var (width, height) = c.GetCurrentSize();
+                c.Resize(width / 2, height);
+            });
+            Render(image2, "Image from file (fit, greyscale, rotated)");
         }
 
         private static void Render(IRenderable canvas, string title)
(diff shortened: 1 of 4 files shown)


src/Spectre.Console/Widgets/Canvas.cs
```diff
@@ -81,6 +84,16 @@ protected override Measurement Measure(RenderContext context, int maxWidth)
 
         /// <inheritdoc/>
         protected override IEnumerable<Segment> Render(RenderContext context, int maxWidth)
+        {
+            return RenderMode switch
```
patriksvensson, line 88: We need to take into account if the console doesn't support Unicode here or is a Windows legacy console, and not use interlaced mode in that case.

```csharp
if(!context.Unicode || context.LegacyConsole) {
   // ...
}
```

src/Spectre.Console/Widgets/Canvas.cs
```diff
@@ -35,17 +35,20 @@ public sealed class Canvas : Renderable
         /// <summary>
         /// Gets or sets the pixel width.
         /// </summary>
-        public int PixelWidth { get; set; } = 2;
+        public int PixelWidth { get; set; } = 1;
```
patriksvensson, line 38: `PixelWidth` should default to `2`

examples/Canvas/Program.cs
```diff
@@ -14,15 +14,31 @@ public static void Main()
...
+            image2.Mutate(ctx => ctx.Grayscale().Rotate(-45).EntropyCrop());
+            image2.Mutate(c =>
+            {
+                var (width, height) = c.GetCurrentSize();
+                c.Resize(width / 2, height);
+            });
+            Render(image2, "Image from file (fit, greyscale, rotated)");
         }
 
         private static void Render(IRenderable canvas, string title)
```
patriksvensson, line 44: Can we leave the example as-is for now?

patriksvensson (changes requested):
A very cool feature. Left a couple of comment that needs to be fixed (together with the warnings and errors) before it can be merged.

Great job!

rogeralsing:
Maybe it would make more sense to have a ICanvasRenderer interface instead of the RenderMode enum?

Where the ICanvasRenderer takes the canvas data and translate it to ascii/unicode depending on its own implementation.

That would open up for 3rd party renderers that don't have to go into the lib itself.

less impact, more flexibility.

It would open up to do things like this:
Skärmavbild 2020-11-30 kl  09 55 48

(From a Java tool, TerminalImageViewer)

Maybe nothing useful but still fun to have the ability to plug such thing in.

patriksvensson:
@rogeralsing Sorry for the late reply.

I think it's OK to leave it as it is right now, and we can create an abstraction if a need for a new renderer arises.

patriksvensson:
@rogeralsing Closing this for now. Feel free to reopen if you want to continue working on this 👍🏻

PR #176, opened 2020-12-23
Adds optional function to get the display string for TextPrompt choices

thoemmi:
Currently, `TextPrompt` uses `TypeConverter`s to display choices. However, it's not always possble to add a TypeConverter, especially if the type originates from a 3rd party library.

This PR allows to specify a custom function to get the display string. For example see the unit test I've added:

```csharp
private class Fruit
{
    public int Id { get; set; }
    public string Text { get; set; }
}

var result = console.Prompt(
    new TextPrompt<Fruit>("Favorite fruit?")
        .AddChoice(new Fruit { Id = 1, Text = "Apple" })
        .AddChoice(new Fruit { Id = 2, Text = "Banana" })
        .WithDisplaySelector(testData => testData.Text));
```

Please note that I've changed `TextPrompt<T>.Choices` from `HashSet` to `List`. This allows the user to order the choices. I assume it's ok that it's the user's responsibility to ensure that the choices are unique.

Last but not least I added an extension method `AddChoices<T>(this TextPrompt<T> obj, IEnumerable<T> choices)`, which enables the user to pass a list of choices instead of adding them individually.


patriksvensson (approved):

patriksvensson:

thoemmi:
Thanks for accepting my PR. I wasn't sure about the `DisplaySelector` name, so I'm glad you've found a better name :wink:

patriksvensson:
It was a solid PR, and makes it much easier to extract selection values! There was nothing wrong with "DisplaySelector", just thought "Converter" was a better fit with the rest of the API. Thanks a lot for this!

PR #181, opened 2020-12-24
Fix nullability in type conversion

SimonCropp:


src/Spectre.Console/Internal/TypeConverterHelper.cs
```diff
@@ -1,55 +1,29 @@
...
             {
                 result = default;
                 return false;
             }
         }
-
-        public static TypeConverter GetTypeConverter<T>()
-        {
-            var converter = TypeDescriptor.GetConverter(typeof(T));
-            if (converter != null)
```
SimonCropp, line 33: TypeDescriptor.GetConverter doesnt return null. so this method is redundant

src/Spectre.Console/Internal/TypeConverterHelper.cs
```diff
@@ -1,55 +1,29 @@
...
-        public static bool TryConvertFromString<T>(string input, [MaybeNull] out T result)
+        public static bool TryConvertFromString<T>(string input, [NotNullWhen(true)] out T? result)
         {
             try
             {
-                result = (T)GetTypeConverter<T>().ConvertFromInvariantString(input);
-                return true;
+                var converter = TypeDescriptor.GetConverter(typeof(T));
+                result = (T?)converter.ConvertFromInvariantString(input);
+                return result != null;
```
SimonCropp, line 20: checking result for null here means we can `NotNullWhen(true)` for the out. so calling methods can use the result without an extra null check

patriksvensson:
@SimonCropp I've rebased this branch against upstream for you, waiting for CI to pass

patriksvensson (approved):
LGTM!

src/Spectre.Console/Internal/TypeConverterHelper.cs
patriksvensson: This method is actually needed. We're not only getting the TypeConverter for the explicit type, but we also check the type if it has a `TypeConverterAttribute` assigned and instantiate that one otherwise.

patriksvensson (changes requested):
Just noticed that the `GetTypeConverter` has been removed which it shouldn't.

SimonCropp:
think it is easier to restart this one

PR #185, opened 2020-12-30
Remove dead code

mcon:


patriksvensson (changes requested):
LGTM! Could you change the commit message to something like `Removed unused TruncateWithEllipsis method` or similar so it's easy to find in case we need to use it again?

patriksvensson (approved):

patriksvensson:
Thanks for this PR! Much appreciated! 🙏

PR #199, opened 2021-01-01
A bit of fun, let's make it easy for people to do cowsay

mcon:
I understand if you'd rather not include this, as it's just a bit of fun!


patriksvensson:
@mcon Although I appreciate it, I don't think this should be included in the library. A great demonstration of how to build custom widgets though!

AdmiringWorm:
@patriksvensson may I suggest perhaps extracting this to the documentation on how to create custom widgets for spectre.console?
I think it may be useful for people wanting to extend with their own widgets, but not knowing how they can create one.

PR #202, opened 2021-01-02
Improvements to tree widget API

patriksvensson:


src/Spectre.Console/Widgets/Tree.cs
```diff
@@ -9,111 +9,150 @@ namespace Spectre.Console
...
         }
 
         /// <inheritdoc />
         protected override IEnumerable<Segment> Render(RenderContext context, int maxWidth)
         {
-            return _rootNode
-                .Render(context, maxWidth)
-                .Concat(new List<Segment> { Segment.LineBreak })
-                .Concat(RenderChildren(context, maxWidth - Rendering.PartSize, _rootNode, depth: 0));
+            if (Nodes.Count == 1)
```
mcon, line 86: Nice spot on this missed case!

src/Spectre.Console/TreeAppearance.cs
```diff
@@ -1,20 +1,22 @@
...
+using Spectre.Console.Rendering;
+
+namespace Spectre.Console
 {
     /// <summary>
-    /// Represents the characters used to render a tree.
+    /// Represents a tree appearance.
     /// </summary>
-    public interface ITreeRendering
+    public abstract partial class TreeAppearance
```
mcon, line 8: Not quite to my tastes, but I guess it is consistent with your existing `<thing>.known` pattern.

mcon (approved):
Nice, makes the API a bunch easier to use.

Worth checking whether the measurement code is actually tested now though :+1:

src/Spectre.Console/TreeAppearance.cs
patriksvensson: Yes, consistency is king (or queen) 😄

patriksvensson:
> Worth checking whether the measurement code is actually tested now though 👍

Yes, absolutely. I will add some more tests for this. The reason I removed them is that measurement tests often become very brittle and hard to reason about when something is changed. I prefer to add tests that render the widget under certain circumstances and makes sure it's correct that way.

patriksvensson:
@mcon Btw, just want to say that these changes in no way disrespect the awesome work that you did. I just wanted to make the API a bit more consistent with other things in the library.

mcon:
Completely understand!

PR #205, opened 2021-01-03
Added canvas unit tests & fix canvas scaling bug

mcon:
No unit tests exist for Canvas, have added some coverage here.

~~Looks like behaviour for rendering a canvas wider than the terminal is broken.~~
Have fixed up `Canvas` to handle the case that it's scaled down too much to be displayed.

### DIFF
--- a/src/Spectre.Console.Tests/Unit/CanvasTests.cs
+++ b/src/Spectre.Console.Tests/Unit/CanvasTests.cs
@@ -27,7 +27,7 @@ namespace Spectre.Console.Tests.Unit
         }
 
         [Fact]
-        public void Render_WiderThan_Terminal()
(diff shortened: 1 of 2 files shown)


patriksvensson:
@mcon Can you elaborate on what's broken?

mcon:
Looks like the resizing behaviour is broken, exposed by Render_Wider_Than_Terminal.

patriksvensson:
> Looks like the resizing behaviour is broken, exposed by Render_Wider_Than_Terminal.

Ah, good catch! If you have the time, feel free to fix it. Otherwise I'll take a look at it later tonight.

patriksvensson (approved):

PR #208, opened 2021-01-04
enables returning a result from StartAsync

thoemmi:
This PR adds overloads to `StartAsync` which accept delegates returning a generic `Task<T>`.

E.g. this allows to show a spinner while fetching data:

```csharp
var data = AnsiConsole.Status()
    .Spinner(Spinner.Known.Star)
    .StartAsync("Fetching data...", async ctx => {
        var r = await _client.FetchDataAsync();
        return r;
    });
```

The code is mainly a copy of the existing functions but returning a generic `Task<T>`.

### DIFF
--- a/src/Spectre.Console/Widgets/Progress/Progress.cs
+++ b/src/Spectre.Console/Widgets/Progress/Progress.cs
@@ -80,34 +80,11 @@ namespace Spectre.Console
                 throw new ArgumentNullException(nameof(action));
             }
 
-            var renderer = CreateRenderer();
-            renderer.Started();
-
-            try
+            _ = await StartAsync<object?>(async progressContext =>
             {
-                using (new RenderHookScope(_console, renderer))
-                {
-                    var context = new ProgressContext(_console, renderer);
-
-                    if (AutoRefresh)
-                    {
-                        using (var thread = new ProgressRefreshThread(context, renderer.RefreshRate))
-                        {
-                            await action(context).ConfigureAwait(false);
-                        }
-                    }
-                    else
-                    {
-                        await action(context).ConfigureAwait(false);
-                    }
-
-                    context.Refresh();
-                }
-            }
-            finally
-            {
-                renderer.Completed(AutoClear);
-            }
+                await action(progressContext);
+                return default;
+            });
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Spectre.Console/Widgets/Progress/Progress.cs
```diff
@@ -110,6 +110,53 @@ public async Task StartAsync(Func<ProgressContext, Task> action)
             }
         }
 
+        /// <summary>
+        /// Starts the progress task list.
+        /// </summary>
+        /// <param name="action">The action to execute.</param>
+        /// <typeparam name="T">The result type of task.</typeparam>
+        /// <returns>A <see cref="Task{T}"/> representing the asynchronous operation.</returns>
+        public async Task<T> StartAsync<T>(Func<ProgressContext, Task<T>> action)
```
patriksvensson, line 119: This method duplicates a lot of code in `Task StartAsync(Func<ProgressContext, Task> action)`. Could we perhaps call this function from the other one with a dummy `object` parameter and discard the result?

src/Spectre.Console/Widgets/Progress/Status.cs
```diff
@@ -85,5 +85,40 @@ public async Task StartAsync(string status, Func<StatusContext, Task> action)
...
         }
+
+        /// <summary>
+        /// Starts a new status display.
+        /// </summary>
+        /// <param name="status">The status to display.</param>
+        /// <param name="action">he action to execute.</param>
+        /// <typeparam name="T">The result type of task.</typeparam>
+        /// <returns>A <see cref="Task{T}"/> representing the asynchronous operation.</returns>
+        public async Task<T> StartAsync<T>(string status, Func<StatusContext, Task<T>> action)
```
patriksvensson, line 96: This method duplicates a lot of code in `Task StartAsync(string status, Func<StatusContext, Task> action)`. Could we perhaps call this function from the other one with a dummy `object` parameter and discard the result?

patriksvensson (changes requested):

src/Spectre.Console/Widgets/Progress/Progress.cs
thoemmi: That bothered me too, but I didn't find a solution. However, after rethinking it again, I think I have resolved the duplication.

patriksvensson (approved):

PR #224, opened 2021-01-13
Add cycle detection to tree rendering

mcon:
In the existing code, cycles will result in an infinite loop in execution, this PR results in an exception being raised in stead.

It's tricky reporting the error in a useful way: i.e. "this is the node that's causing problems", given that would involve rendering the node, and including that information in the exception rather than printing to console, would involve using a fake console, which would be all very strange. Certainly this is better than the previous state of affairs, let me know if you have any ideas here.


patriksvensson:
@mcon Ready for review?

src/Spectre.Console.Tests/Unit/TreeTests.cs
```diff
@@ -56,5 +57,26 @@ public Task Should_Render_Tree_With_No_Child_Nodes_Correctly()
...
+            child1.AddNodes(child2, child3);
+            var root = new TreeNode(new Text("Branch Node"));
+            root.AddNodes(child1);
+            child2.AddNode(root);
+
+            var tree = new Tree("root node");
+            tree.AddNodes(root);
+
+            // When
+            Should.Throw<CircularTreeException>(() => console.Render(tree));
```
patriksvensson, line 79:
```suggestion
            // When
            var result = Record.Exception(() => console.Render(tree));
            
            // Then
            result.ShouldBeOfType<CircularTreeException>();
```

src/Spectre.Console.Tests/Unit/TreeTests.cs
```diff
@@ -56,5 +57,26 @@ public Task Should_Render_Tree_With_No_Child_Nodes_Correctly()
             // Then
             return Verifier.Verify(console.Output);
         }
+
+        [Fact]
+        public void Tree_With_Cycle_Throws()
```
patriksvensson, line 62:
```suggestion
        public void Should_Throw_If_Tree_Contains_Cycles()
```

patriksvensson (changes requested):

patriksvensson (approved):

patriksvensson:
@mcon There are merge conflicts in the PR branch. Could you resolve them and I'll merge. Thanks 👍🏻

mcon:
Doesn't look like there are any conflicts to me...

patriksvensson:
Ah, couldn't rebase and merge due to conflict in individual commits. Squashing went fine.

patriksvensson:

PR #228, opened 2021-01-13
mask default value when prompt is a secret

thoemmi:
Currently the default value is displayed even if the text prompt is a secret. This PR will mask it with `******`.

(First I displayed the same number of `*` as the length of the default value; however, to prevent exposing the length of the default value, I replaced it with a fixed length of 6)


patriksvensson (approved):

patriksvensson:

PR #236, opened 2021-01-15
Add support for default choice in selection prompt

### LINKED ISSUE #234: Add a way to mark items as default (MultiSelectionPrompt`1)
Currently, it is not possible to have items that are marked/selected by default.

```cs
var prompt = new MultiSelectionPrompt<string>()
     .Selected(1, 2, 3, 4)
     .AddChoices(...);
```

This would produce the following (without having any user input):

```bash
...
 [ ] a
 [X] b
 [X] c
 [X] d
 [X] e
 [ ] f
...
```

Using a hilarious construct of expression trees that add an intercepting render pipeline that catches the prompt's render hook and get the selections of the renderable list abstraction to mutate the internal HashSet - which is my current solution... but why would anyone do that?

**Additional context**
\-

patriksvensson:
Closes #234

### DIFF
--- a/src/Spectre.Console/Widgets/Prompt/MultiSelectionPrompt.cs
+++ b/src/Spectre.Console/Widgets/Prompt/MultiSelectionPrompt.cs
@@ -24,6 +24,11 @@ namespace Spectre.Console
         public List<T> Choices { get; }
 
         /// <summary>
+        /// Gets the initially selected choices.
+        /// </summary>
+        public HashSet<int> Selected { get; }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


thoemmi:
Wouldn't be a list of values instead of indices a better way to pre-select items,? 🤔 E.g.
```csharp
var prompt = new MultiSelectionPrompt<string>()
     .AddChoices("apple", "banana", "orange");
     .Selected("apple", "banana");
```

patriksvensson:
Yeah, maybe. Went with that as well first but there are some gotchas API wise with custom types. Easy to add an extension method though if someone needs to pass the actual item instances.

PR #245, opened 2021-01-23
SetApplicatiionVersion / better was to auto-detect application version

MiloszKrajewski:
* added SetApplicationVersion(...) to configurator, so version can be overridden (but doesn't have to)
* changed VersionHelper to retrieve assembly version in a way compatible with single file executable
* modified "Demo" application to actually have a version included in csproj and/or configured from Program.cs
* ApplicationName was also falling back to Assembly.Location which does not work in single file executable

### DIFF
--- a/examples/Cli/Demo/Demo.csproj
+++ b/examples/Cli/Demo/Demo.csproj
@@ -8,6 +8,7 @@
     <ExampleDescription>Demonstrates the most common use cases of Spectre.Cli.</ExampleDescription>
     <ExampleGroup>Cli</ExampleGroup>
     <ExampleVisible>false</ExampleVisible>
+    <Version>1.2.3.4-beta.5</Version>
   </PropertyGroup>
 
   <ItemGroup>
--- a/examples/Cli/Demo/Program.cs
+++ b/examples/Cli/Demo/Program.cs
@@ -11,6 +11,7 @@ namespace Demo
             app.Configure(config =>
             {
                 config.SetApplicationName("fake-dotnet");
+                // config.SetApplicationVersion("1.3.3.7-ultra!");
                 config.ValidateExamples();
                 config.AddExample(new[] { "run", "--no-build" });
 
--- a/src/Spectre.Console/Cli/CommandApp.cs
+++ b/src/Spectre.Console/Cli/CommandApp.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading.Tasks;
 using Spectre.Console.Rendering;
 
@@ -84,7 +85,7 @@ namespace Spectre.Console.Cli
                 }
 
                 return await _executor
-                    .Execute(_configurator, args)
+                    .Execute(_configurator, args.ToArray())
                     .ConfigureAwait(false);
             }
             catch (Exception ex)
--- a/src/Spectre.Console/Cli/Internal/CommandExecutor.cs
+++ b/src/Spectre.Console/Cli/Internal/CommandExecutor.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Reflection;
 using System.Threading.Tasks;
 
 namespace Spectre.Console.Cli
@@ -16,7 +15,7 @@ namespace Spectre.Console.Cli
             _registrar.Register(typeof(DefaultPairDeconstructor), typeof(DefaultPairDeconstructor));
         }
 
-        public Task<int> Execute(IConfiguration configuration, IEnumerable<string> args)
+        public Task<int> Execute(IConfiguration configuration, IReadOnlyCollection<string> args)
         {
             if (configuration == null)
             {
@@ -44,7 +43,7 @@ namespace Spectre.Console.Cli
                         firstArgument.Equals("-v", StringComparison.OrdinalIgnoreCase))
                     {
                         var console = configuration.Settings.Console.GetConsole();
-                        console.WriteLine(VersionHelper.GetVersion(Assembly.GetEntryAssembly()));
+                        console.WriteLine(ResolveApplicationVersion(configuration));
                         return Task.FromResult(0);
                     }
                 }
@@ -82,6 +81,10 @@ namespace Spectre.Console.Cli
             // Execute the command tree.
             return Execute(leaf, parsedResult.Tree, context, resolver, configuration);
         }
+
+        private static string ResolveApplicationVersion(IConfiguration configuration) =>
+            configuration.Settings.ApplicationVersion ?? // potential override
+            VersionHelper.GetApplicationVersion();
 
         private static Task<int> Execute(
             CommandTree leaf,
--- a/src/Spectre.Console/Cli/Internal/Modelling/CommandModel.cs
+++ b/src/Spectre.Console/Cli/Internal/Modelling/CommandModel.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Reflection;
 
@@ -26,9 +27,16 @@ namespace Spectre.Console.Cli
             Examples = new List<string[]>(examples ?? Array.Empty<string[]>());
         }
 
-        public string GetApplicationName()
-        {
-            return ApplicationName ?? Path.GetFileName(Assembly.GetEntryAssembly()?.Location) ?? "?";
-        }
+        public string GetApplicationName() =>
+            ApplicationName ??
+            Path.GetFileName(GetApplicationFile()) ?? // it is actually null safe
+            "?";
+
+        private static string? GetApplicationFile() =>
+            Assembly.GetEntryAssembly()?.Location switch // location can be empty string or null
+            {
+                var location when !string.IsNullOrWhiteSpace(location) => location,
+                _ => Process.GetCurrentProcess().MainModule?.FileName,
+            };
     }
 }
--- a/src/Spectre.Console/Cli/Internal/VersionHelper.cs
+++ b/src/Spectre.Console/Cli/Internal/VersionHelper.cs
@@ -5,22 +5,20 @@ namespace Spectre.Console.Cli
 {
     internal static class VersionHelper
     {
-        public static string GetVersion(Assembly? assembly)
-        {
-            if (assembly == null)
-            {
-                return "?";
-            }
+        private static string? TryGetAssemblyVersion(Assembly? assembly) =>
+            assembly?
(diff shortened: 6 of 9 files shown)


examples/Cli/Demo/Program.cs
```diff
@@ -11,6 +11,7 @@ public static int Main(string[] args)
             app.Configure(config =>
             {
                 config.SetApplicationName("fake-dotnet");
+                // config.SetApplicationVersion("1.3.3.7-ultra!");
```
patriksvensson, line 14: This should be removed?

examples/Cli/Demo/Demo.csproj
```diff
@@ -8,6 +8,7 @@
     <ExampleDescription>Demonstrates the most common use cases of Spectre.Cli.</ExampleDescription>
     <ExampleGroup>Cli</ExampleGroup>
     <ExampleVisible>false</ExampleVisible>
+    <Version>1.2.3.4-beta.5</Version>
```
patriksvensson, line 11: Perhaps remove this since version is autopopulated by MinVer.

src/Spectre.Console/Cli/CommandApp.cs
```diff
@@ -84,7 +85,7 @@ public async Task<int> RunAsync(IEnumerable<string> args)
                 }
 
                 return await _executor
-                    .Execute(_configurator, args)
+                    .Execute(_configurator, args.ToArray())
```
patriksvensson, line 88: Why `.ToArray` here?

src/Spectre.Console/Cli/Internal/CommandExecutor.cs
```diff
@@ -16,7 +15,7 @@ public CommandExecutor(ITypeRegistrar registrar)
             _registrar.Register(typeof(DefaultPairDeconstructor), typeof(DefaultPairDeconstructor));
         }
 
-        public Task<int> Execute(IConfiguration configuration, IEnumerable<string> args)
+        public Task<int> Execute(IConfiguration configuration, IReadOnlyCollection<string> args)
```
patriksvensson, line 18: Why has this been changed to a `IReadOnlyCollection<string>`?

src/Spectre.Console/Cli/Internal/VersionHelper.cs
```diff
@@ -5,22 +5,20 @@ namespace Spectre.Console.Cli
...
+        private static FileVersionInfo? TryGetMainModuleVersionInfo() =>
+            Process.GetCurrentProcess().MainModule?.FileVersionInfo;
+
+        public static string GetVersion(Assembly? assembly) =>
+            TryGetAssemblyVersion(assembly) ?? "?";
+
+        public static string GetApplicationVersion() =>
+            TryGetAssemblyVersion(Assembly.GetEntryAssembly()) ??
+            TryGetMainModuleVersionInfo()?.ProductVersion ??
+            "?";
```
patriksvensson, line 22: What's the difference between `GetVersion` and `GetApplicationVersion`? I see that both methods are used at different places. 

GetVersion: This one is used in `VersionCommand`
GetApplicationVersion: This one is used in `GetApplicationVersion`

src/Spectre.Console/Cli/Internal/VersionHelper.cs
```diff
@@ -5,22 +5,20 @@ namespace Spectre.Console.Cli
 {
     internal static class VersionHelper
```
patriksvensson, line 6: Would prefer if you used brackets around methods and explicit usage of the `return` keyword.

src/Spectre.Console/Cli/Internal/Modelling/CommandModel.cs
```diff
@@ -26,9 +27,16 @@ internal sealed class CommandModel : ICommandContainer
...
-        {
-            return ApplicationName ?? Path.GetFileName(Assembly.GetEntryAssembly()?.Location) ?? "?";
-        }
+        public string GetApplicationName() =>
+            ApplicationName ??
+            Path.GetFileName(GetApplicationFile()) ?? // it is actually null safe
+            "?";
+
+        private static string? GetApplicationFile() =>
+            Assembly.GetEntryAssembly()?.Location switch // location can be empty string or null
```
patriksvensson, line 36: Can we change this to a simple if-else statement?

src/Spectre.Console/Cli/Internal/Modelling/CommandModel.cs
```diff
@@ -26,9 +27,16 @@ internal sealed class CommandModel : ICommandContainer
             Examples = new List<string[]>(examples ?? Array.Empty<string[]>());
         }
 
-        public string GetApplicationName()
-        {
-            return ApplicationName ?? Path.GetFileName(Assembly.GetEntryAssembly()?.Location) ?? "?";
-        }
+        public string GetApplicationName() =>
```
patriksvensson, line 30: Would prefer if you used brackets around methods and explicit usage of the return keyword.

src/Spectre.Console/Cli/Internal/CommandExecutor.cs
```diff
@@ -83,6 +82,10 @@ public Task<int> Execute(IConfiguration configuration, IEnumerable<string> args)
             return Execute(leaf, parsedResult.Tree, context, resolver, configuration);
         }
 
+        private static string ResolveApplicationVersion(IConfiguration configuration) =>
```
patriksvensson, line 85: Would prefer if you used brackets around methods and explicit usage of the return keyword.

patriksvensson (changes requested):

src/Spectre.Console/Cli/CommandApp.cs
MiloszKrajewski: I guess it is an OCD. Inside `Execute` `IEnumerable<T>` is enumerated twice, which is a little smell and as such is highlighted by Rider/ReSharper. I've changed it to `IReadOnlyCollection<T>` to avoid double enumeration inside but it had to become cached on the outside (thus `.ToArray()`)
If you don't like it I can change it back, there is no real impact on double enumeration in this particular case I just have this OCD-like reaction to Rider/ReSharper suggestions/warnings.

src/Spectre.Console/Cli/Internal/CommandExecutor.cs
MiloszKrajewski: Is the explanation above clear? (I have tendency to ramble)

src/Spectre.Console/Cli/Internal/VersionHelper.cs
MiloszKrajewski: `GetVersion` is taking version of ANY assembly, the other one is taking version of application (usually EntryAssembly, but has a special case for `Process.GetCurrentProcess().MainModule?.FileVersionInfo` which has nothing to do with assemblies).

You use them in two different contexts: one is taking version of `Spectre.Console` in particular, while the other one is taking version of `.exe` which is running ("the host of Spectre.Console").

examples/Cli/Demo/Program.cs
MiloszKrajewski: I used it to test if overriding works. You cannot really write unit tests for this problem because problem is very specifically related to "publishing for win-x64 as single file executable with .NET 5" which is a little bit deployment thing rather than unit test.

I left it kind of for you, so you can check if and how it works.
Should I remove?

examples/Cli/Demo/Demo.csproj
MiloszKrajewski: I didn't know how this (MinVer) works, but I used this for testing (again) so I'm sure I'm taking host app version not some assembly inside version, that's why this number is very distinctive.

Again, can be removed it was just for me to show you how/if it works.

src/Spectre.Console/Cli/Internal/Modelling/CommandModel.cs
MiloszKrajewski: If you have bracket that it can. with 'if' it was just more-than-one statements, and I like expression-bodied-functions.
But yeah, sure, can be proper multi statement function.

src/Spectre.Console/Cli/Internal/VersionHelper.cs
MiloszKrajewski: LOL. So there will be few of them then ;-)
My personal preferences are completely opposite, but yes - no problem. I'll change them all.

src/Spectre.Console/Cli/Internal/Modelling/CommandModel.cs
MiloszKrajewski: Yup. See above.

src/Spectre.Console/Cli/Internal/CommandExecutor.cs
MiloszKrajewski: Yup. See above.

examples/Cli/Demo/Demo.csproj
patriksvensson: Yes, remove this.

examples/Cli/Demo/Program.cs
patriksvensson: Yes, remove this.

src/Spectre.Console/Cli/CommandApp.cs
patriksvensson: I understand, but I think the enumerable should be materialized inside the Execute if it's enumerated twice.

src/Spectre.Console/Cli/Internal/CommandExecutor.cs
patriksvensson: Yes. Can we change this back to an `IEnumerable<string>`? You can create a private overload that accepts an `IReadOnlyList<string>` if you want to and call that one from the public method.

patriksvensson (changes requested):

src/Spectre.Console/Cli/Internal/CommandExecutor.cs
MiloszKrajewski: Absolutely yes. I wasn't thinking about it as "public" interface... because I never used it ;-) but it absolutely makes sense to keep it as IEnumerable on public interface.

MiloszKrajewski: I revered it to `IEnumerable<T>`. Although there is chance for double enumeration but it would actually affect performance only if given enumerable had high startup cost, and in 99,9% of cases it is actually an array (or Segment/Slice) so it doesn't matter. If you can live with ReSharper warning, I can suppress my OCD ;)

patriksvensson:
@MiloszKrajewski Merged! 👍 Thank you a lot for this, much appreciated! 🙏

PR #253, opened 2021-02-01
Add formatting support for breakdown chart values

### LINKED ISSUE #252: Formatting for tag value in breakdown chart
Be nice if the tag's value could have a custom format. My simple example I'd like to just do a `{0:N}` format on them to get commas, but i could see someone wanting to use something like `4MB`, `49 seconds`, etc

Maybe something like this?

```csharp
var chart = new BreakdownChart()
    .TagValueFormat(value => $"{value:N}")
    .AddItems(insertReport,
        pair => new BreakdownChartItem(pair.Key, pair.Value, counter++)
    );
```

**Additional context**
I could knock this out in a PR pretty easily if needed.  Nothing pressing, just thought it'd be neat to have.

patriksvensson:
Closes #252


examples/Console/Charts/Program.cs
```diff
@@ -22,7 +22,6 @@ public static void Main()
             Render("Languages", new BreakdownChart()
                 .FullSize()
                 .Width(60)
```
patriksvensson, line 24: Format tag values as percent here.

PR #255, opened 2021-02-11
Treat stderr as equivalent to stdout

bastianeicher:
I would like to use Spectre.Console to write certain output like progress bars to stderr so that it remains visible even when stdout is redirected. I hoped to achieve this with:

```
var stderrConsole = AnsiConsole.Create(new AnsiConsoleSettings {Out = Console.Error});
```

However the ANSI control characters don't seem to work on stderr. I think this may be because of checks like these:

https://github.com/spectresystems/spectre.console/blob/e20f6284f962d53eab725285c76303dac768f62a/src/Spectre.Console/AnsiConsoleFactory.cs#L30-L31

I'm hoping that the change in this PR will cause both stdout and stderr to be treated as "normal" non-redirected console output.

### DIFF
--- a/src/Spectre.Console/AnsiConsoleFactory.cs
+++ b/src/Spectre.Console/AnsiConsoleFactory.cs
@@ -28,7 +28,7 @@ namespace Spectre.Console
             var (supportsAnsi, legacyConsole) = DetectAnsi(settings, buffer);
 
             // Use the provided encoding or fall back to UTF-8
-            var encoding = buffer.IsStandardOut() ? System.Console.OutputEncoding : Encoding.UTF8;
+            var encoding = buffer.IsStandardOut() || buffer.IsStandardError() ? System.Console.OutputEncoding : Encoding.UTF8;
 
             // Get the color system
             var colorSystem = settings.ColorSystem == ColorSystemSupport.Detect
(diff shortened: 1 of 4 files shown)


patriksvensson:
@bastianeicher I'm sorry but I can't merge this PR since it will mess up some other internals in Spectre.Console and will change the meaning of `IsStandardOut()` which in turn certainly will cause some bugs in the future.

The reason why ANSI escape codes aren't being written not have anything to do with the encoding, but whether or not the current terminal supports ANSI escape codes. What do the values in `IAnsiConsole.Profile.Capabilities` look like?

bastianeicher:
I ran the following test code with .NET 5.0:

```csharp
AnsiConsole.WriteLine("stdout");
AnsiConsole.Progress().Start(ctx => ctx.AddTask("My Task"));

var stderr = AnsiConsole.Create(new AnsiConsoleSettings {Out = System.Console.Error});
stderr.WriteLine("stderr");
stderr.Progress().Start(ctx => ctx.AddTask("My Task"));
```

On Windows 10 20H2 the output looks like this:
encoding-error

With my PR the output looks like this:
encoding-ok

`.Capabilities` is `ANSI=Yes, Colors=TrueColor, Kind=Modern (24 bits)` for both `AnsiConsole` and `stderr`.

Changing this line:
https://github.com/spectresystems/spectre.console/blob/e20f6284f962d53eab725285c76303dac768f62a/src/Spectre.Console/AnsiConsoleFactory.cs#L31
to
```csharp
var encoding = System.Console.OutputEncoding;
```
also "fixed" the issue (though it's obviously not the right solution). This leads me to believe that the issue does indeed have to do with the encoding.

patriksvensson:
@bastianeicher Have you tried setting the encoding explicitly?

```
AnsiConsole.Profile.Encoding = System.Console.OutputEncoding
```

I'd rather change that line to something like the one below if we need to change it.

```csharp
var encoding = buffer.IsStandardOut() || buffer.IsStandardErr() ? System.Console.OutputEncoding : Encoding.UTF8; 
```

bastianeicher:
I've updated the PR to introduce a new `IsStandardError()` extension method and use it where I thought it appropriate.

patriksvensson (approved):

patriksvensson:

PR #258, opened 2021-02-14
Adds text and Progress bar spinner column for tasks yet to be started

phil-scott-78:
I was running into this issue. 

image

My code is setting `CompletedText` to the check box, but then I started setting `AutoStart` to false to get better results for things like download speed and noticed the checkbox was popping up for tasks yet to be started. 

For backwards compatibility, `PendingText` should probably fall back to `CompletedText` but I suspect I'm the only one using this feature at this point so probably no need to introduce that complexity. Hopefully.

### DIFF
--- a/src/Spectre.Console/Widgets/Progress/Columns/SpinnerColumn.cs
+++ b/src/Spectre.Console/Widgets/Progress/Columns/SpinnerColumn.cs
@@ -53,7 +53,7 @@ namespace Spectre.Console
 
         /// <summary>
         /// Gets or sets the text that should be shown instead
-        /// of the spinner once a task completes.
+        /// of the spinner before a task begins.
         /// </summary>
         public string? PendingText
         {
@@ -71,7 +71,7 @@ namespace Spectre.Console
         public Style? CompletedStyle { get; set; }
 
         /// <summary>
-        /// Gets or sets the completed style.
+        /// Gets or sets the pending style.
         /// </summary>
         public Style? PendingStyle { get; set; }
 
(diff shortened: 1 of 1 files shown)


src/Spectre.Console/Widgets/Progress/Columns/SpinnerColumn.cs
```diff
@@ -50,11 +51,30 @@ public Spinner Spinner
...
+            }
+        }
+
         /// <summary>
         /// Gets or sets the completed style.
         /// </summary>
         public Style? CompletedStyle { get; set; }
 
+        /// <summary>
+        /// Gets or sets the completed style.
```
patriksvensson, line 74:
```suggestion
        /// Gets or sets the pending style.
```

src/Spectre.Console/Widgets/Progress/Columns/SpinnerColumn.cs
```diff
@@ -50,11 +51,30 @@ public Spinner Spinner
             }
         }
 
+        /// <summary>
+        /// Gets or sets the text that should be shown instead
+        /// of the spinner once a task completes.
```
patriksvensson, line 56: This text hasn't been updated after copy+paste.

patriksvensson (changes requested):

phil-scott-78:
My copy and paste skills aren't weren't they used to be. Updated.

patriksvensson (approved):

patriksvensson:

PR #275, opened 2021-02-27
using loop instead of linq to reduce allocs

phil-scott-78:
In both loops, the context is captured preventing caching of the lambda. this results in a pretty significant number of allocations especially with progress bars that constantly are remeasuring

I noticed this while working on the update panel. over a 2-minute-long progress I was seeing a half gig worth of allocations there. After measuring it I noticed a huge portion of the allocations come from these two `Sum` functions. To keep it simple for this I just measured the allocs for the progress example

## before
before

## after
after

if you check the right-hand side you can see 9mb worth of allocations are gone for even just a 6s run.  For the 2-minute-long version run of the update panel it cut about 320mb out of 500mb.

### DIFF
--- a/src/Spectre.Console/Rendering/Segment.cs
+++ b/src/Spectre.Console/Rendering/Segment.cs
@@ -135,7 +135,13 @@ namespace Spectre.Console.Rendering
                 throw new ArgumentNullException(nameof(segments));
             }
 
-            return segments.Sum(segment => segment.CellCount(context));
+            var sum = 0;
+            foreach (var segment in segments)
+            {
+                sum += segment.CellCount(context);
+            }
+
+            return sum;
         }
 
         /// <summary>
--- a/src/Spectre.Console/Internal/Cell.cs
+++ b/src/Spectre.Console/Internal/Cell.cs
@@ -8,7 +8,13 @@ namespace Spectre.Console
     {
         public static int GetCellLength(RenderContext context, string text)
         {
-            return text.Sum(rune => GetCellLength(context, rune));
+            var sum = 0;
+            foreach (var rune in text)
+            {
+                sum += GetCellLength(context, rune);
+            }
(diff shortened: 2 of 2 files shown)


src/Spectre.Console/Rendering/Segment.cs
```diff
@@ -79,7 +79,7 @@ public Segment(string text, Style style)
 
         private Segment(string text, Style style, bool lineBreak, bool control)
         {
-            Text = text?.NormalizeNewLines() ?? throw new ArgumentNullException(nameof(text));
+            Text = text.NormalizeNewLines() ?? throw new ArgumentNullException(nameof(text));
```
patriksvensson, line 82: I think we still want the `text?` part since a `text` that is null will throw a null reference exception here.

```suggestion
            Text = text?.NormalizeNewLines() ?? throw new ArgumentNullException(nameof(text));
```

patriksvensson (changes requested):
Awesome! Performance is something we really need to look into and this is a great first step in reducing allocations! 👍🏻 

Found one minor thing that I think needs to be fixed though.

src/Spectre.Console/Rendering/Segment.cs
phil-scott-78: text would have had to be defined as `string?` to require a null check here, right?  But that change doesn't belong in this commit regardless. Ripped it out plus a whitespace change that snuck in.

patriksvensson: @phil-scott-78 Correct, but since Segment is a public API, we can't guarantee that someone doesn't call the method with a null string if nullable reference types are turned off.

patriksvensson (approved):

patriksvensson:
@phil-scott-78 Merged! 👍

PR #276, opened 2021-02-28
Support setting the static console

patriksvensson:
This PR makes it possible to set/replace the current static console.


phil-scott-78 (approved):
LGM

PR #277, opened 2021-03-01
Progress remaining time to show elapsed when finished

MarkPflug:
Currently, the progress bar `RemainingTimeColumn` will render as "00:00:00" when the task is finished. It would be nice to have this render the total elapsed time instead once the task is finished. I know I could add the elapsed time column as well, but with both remaining and elapsed displayed it becomes a bit overwhelming visually. This collapses the function of both columns into one.

It is entirely opt-in, and isn't a breaking change. Existing callers will continue to get the same behavior.

### DIFF
--- a/src/Spectre.Console/Widgets/Progress/Columns/HybridTimeColumn.cs
+++ b/src/Spectre.Console/Widgets/Progress/Columns/HybridTimeColumn.cs
@@ -1,0 +1,62 @@
+using System;
+using Spectre.Console.Rendering;
+
+namespace Spectre.Console
+{
+    /// <summary>
+    /// A column showing the remaining time of a task, and elapsed time upon completion.
+    /// </summary>
+    public sealed class HybridTimeColumn : ProgressColumn
+    {
+        /// <inheritdoc/>
+        protected internal override bool NoWrap => true;
+
+        /// <summary>
+        /// Gets or sets the style of the remaining time text.
+        /// </summary>
+        public Style Style { get; set; } = new Style(foreground: Color.Blue);
+
+        /// <summary>
+        /// Gets or sets the text style to render the elapsed time on completion.
+        /// </summary>
+        public Style FinishedStyle { get; set; } = new Style(foreground: Color.Green);
+
+        /// <inheritdoc/>
+        public override IRenderable Render(RenderContext context, ProgressTask task, TimeSpan deltaTime)
+        {
+            if (task.IsFinished)
+            {
+                var elapsedTime = task.ElapsedTime;
+                if (elapsedTime == null)
+                {
+                    return new Markup("--:--:--");
+                }
+                else
+                {
+                    return new Text($"{elapsedTime.Value:hh\\:mm\\:ss}", FinishedStyle ?? Style.Plain);
+                }
+            }
+            else
+            {
+                var remaining = task.RemainingTime;
+                if (remaining == null)
+                {
+                    return new Markup("--:--:--");
+                }
+
+                if (remaining.Value.TotalHours > 99)
+                {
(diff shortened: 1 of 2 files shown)


src/Spectre.Console/Widgets/Progress/Columns/RemainingTimeColumn.cs
```diff
@@ -8,6 +8,25 @@ namespace Spectre.Console
...
+        public RemainingTimeColumn()
+            : this(false)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RemainingTimeColumn"/> class.
+        /// </summary>
+        /// <param name="elapsedWhenFinished">Indicates if the elapsed time should display when the task is complete.</param>
+        public RemainingTimeColumn(bool elapsedWhenFinished)
```
phil-scott-78, line 23: Just to match the style of the other columns it'd be better to just drop the constructors and let this be set via a property

src/Spectre.Console/Widgets/Progress/Columns/RemainingTimeColumn.cs
```diff
@@ -16,21 +35,41 @@ public sealed class RemainingTimeColumn : ProgressColumn
         /// </summary>
         public Style Style { get; set; } = new Style(foreground: Color.Blue);
 
+        /// <summary>
+        /// Gets or sets the text style to render the elapsed time on completion.
+        /// </summary>
+        public Style FinishedStyle { get; set; } = new Style(foreground: Color.Green);
```
phil-scott-78, line 41: This name implies to me that this would the style used for all scenarios when `IsFinished` is true, not just when `elapsedWhenFinished` is true. I don't think it needs to be changed, but I'd adjust the render code to take that into account

MarkPflug: The thought occurred to me too, I didn't apply it to both code paths because I didn't want to perturb any existing callers who might find the change of color surprising.

MarkPflug:
Perhaps, as an alternative design: leave RemainingTimeColumn as-is, and add a new HybridTimeColumn to implement this behavior instead? Sound reasonable? The hybrid time column could then apply colors more consistently without worrying about changing colors for existing users.

phil-scott-78:
I like that idea of a new column. A little more explicit about its purpose.

src/Spectre.Console/Widgets/Progress/Columns/HybridTimeColumn.cs
```diff
@@ -0,0 +1,62 @@
+using System;
+using Spectre.Console.Rendering;
+
+namespace Spectre.Console
+{
+    /// <summary>
+    /// A column showing the remaining time of a task, and elapsed time upon completion.
+    /// </summary>
+    public sealed class HybridTimeColumn : ProgressColumn
```
patriksvensson, line 9: Perhaps we should try and find another word than `Hybrid` here that is a little more descriptive of what it does.

MarkPflug: How about TaskTimerColumn or TaskTimingColumn? Could also go back to adding as a property on RemainingTimeColumn instead.
Open to suggestions, naming is hard...

patriksvensson:
@MarkPflug I've been having some discussions with @phil-scott-78 about what would be the best approach forward when it comes to customizing columns, and we are still debating how to make columns more extensible. Can we hold off on this PR while we think about how to get the API right?

In the meanwhile, perhaps you can use the code in this PR as a workaround?

Issue #279, opened 2021-03-03
ProgressTask.Stop() does not work properly


Yaroshvitaliy:
ProgressTask.Stop() does not stop progress task completely, it remain be running (spinner is active). The only way to finish the task is to set 100% progress that is not acceptable for my task. And the actual problem is that this progress is being rendered over the rest UI, including other instances of progresses, that makes the following issue critical:
https://github.com/spectresystems/spectre.console/issues/259

phil-scott-78:
I agree, that does seem like unexpected behavior. Without context on why you are stopping the tasks I want to clarify a couple of things with you before we start planning a fix. 

Currently a lot of assumptions are made about a task. It is either waiting to be started, running, or finished. Once a task is finished you can't return to running or started. This is important because it not only drives UI (e.g. like you are seeing with the spinner), but also for concepts like when a Progress bar itself knows when all its children are completed so it knows when to finish itself.  There is no concept of a paused task beyond just a progress bar that isn't actively getting its `Increment` method called

So, for this bug I'd propose a couple of changes to `ProgressTask`

1. Change `IsFinished` to `StopTime != null || Value >= MaxValue`. 
2. Throw an exception on `StartTask` if `StopTime != null`. Without this being explicit 
3. Add documentation to `StopTask` indicating that it can't be started again

This would obviously be a breaking behavior, but I suspect with no one having submitted this bug before I think it's a safe assumption not a lot of people are stopping and starting tasks because they surely would have noticed this bug.

Yaroshvitaliy:
Thank you for your reply. 
I agree that `IsFinished ` should consider `StopTime != null`.
I think no one has submitted this bug before simply because it's realy rare case when it's needed to stop ptogress task.

I've found such a workaround as a temporary solution for a stopped progress task.
1. increment the progress task to set progress 100% - this will stop the task 
2. set progress' task description like 'stopped: {task.Percentage}%'

This is not an ideal solution, but it's critical to me to show when progress task was stopped.

bddckr:
In my particular case I have a task that can actually be skipped altogether. As it's one task of many, I still want to have it in the progress' tasks list - visible to the user. In this sense I believe my requirements match @Yaroshvitaliy's: We need the ability to stop a task and freeze it's progress to ensure we do not lose the percentage information.

[...]

PR #282, opened 2021-03-03
Add Progress.HideCompleted

### LINKED ISSUE #281: Progress to remove tasks as they are completed
I have a very long (up to 60+) list of tasks which I'd like to show progress for. They are not all added at the beginning, and some complete before new ones are added. I'd say the number of the started but non-completed tasks should not be larger then 20 at any time. Therefore I'd like an option to remove completed tasks from the display to keep the list visible. Consider that in addition to the progress bars there are going to be status messages displayed, just like in the Progress example.

A simple change in `DefaultProgressRenderer.cs` would give me what I need:
foreach (var task in context.GetTasks().**Where(tsk => !tsk.IsFinished)**)

Of course, the filtering should be controlled by an option (how about "HideCompleted(bool)", or "ClearCompleted(bool)"?) with the same syntax as "AutoClear". If you think this is a good idea, I can try and work on a Pull Request. The one thing I am not comfortable with is providing proper unit tests for the new feature for which I would need some guidance or hints. I noticed there seems to be only one test around AutoClear(true) (`Should_Render_Task_Correctly`). Something like that should be within my capabilities :-)

lstefano71:
Closes #281

### DIFF
--- a/docs/input/progress.md
+++ b/docs/input/progress.md
@@ -63,6 +63,7 @@ await AnsiConsole.Progress()
 AnsiConsole.Progress()
     .AutoRefresh(false) // Turn off auto refresh
     .AutoClear(false)   // Do not remove the task list when done
+    .HideCompleted(false)   // Hide tasks as they are completed
     .Columns(new ProgressColumn[] 
     {
         new TaskDescriptionColumn(),    // Task description
--- a/src/Spectre.Console.Tests/Unit/Progress/ProgressTests.cs
+++ b/src/Spectre.Console.Tests/Unit/Progress/ProgressTests.cs
@@ -189,7 +189,9 @@ namespace Spectre.Console.Tests.Unit
         public void Should_Hide_Completed_Tasks()
         {
             // Given
-            var task = default(ProgressTask);
+            var taskFinished = default(ProgressTask);
+            var taskInProgress1 = default(ProgressTask);
+            var taskInProgress2 = default(ProgressTask);
             var console = new FakeAnsiConsole(ColorSystem.TrueColor, width: 10);
 
             var progress = new Progress(console)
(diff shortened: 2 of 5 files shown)


phil-scott-78:
This looks like a first-class PR to me. The build error looks like some sort of nuget restore issue.  I suspect if we rerun the build that it'll resolve itself. @patriksvensson, any concerns with manual starts like that? I haven't dug into the CI too much 

The one thing that would be nice to see would be the unit test verify the display with one task complete and one incomplete, but I'm not 100% sure if that's possible until the `StopTask()` #279 is resolved.

patriksvensson:
> @patriksvensson, any concerns with manual starts like that? I haven't dug into the CI too much

No concerns. I've restarted the CI for you 😊

lstefano71:
Thank you for your encouraging words. 
A look at the details of the build log for the MacOS version showed me that it must have been a fluke of the CI platform so I wasn't too worried about that.

> The one thing that would be nice to see would be the unit test verify the display with one task complete and one incomplete, but I'm not 100% sure if that's possible until the `StopTask()` #279 is resolved.

Indeed! I came to the same conclusion immediately after submitting the PR but I was already heading to bed for the day and while I am used to do coding in my pajamas in the wee hours of the morning, these days I need rest more than ever. I am about to submit a commit which tries to prove that a completed task will be hidden while tasks in progress still show up in the Console.

lstefano71:
> I am about to submit a commit which tries to prove that a completed task will be hidden while tasks in progress still show up in the Console.

Improved unit test: bc4846028559

lstefano71:
I found the source for the documentation and I thought I'd mention the new property in the "Configure" section of the page dedicated to the Progress widget in commit 7b207d9 to make it more discoverable by users like me who like to browse the docs.

phil-scott-78 (approved):

PR #283, opened 2021-03-04
Serilog example for logging

phil-scott-78:
I tried to keep this as simple as a solution for a sneakily tricky problem like this can be.

Normally I'm not sure having examples tied to specific libraries would be an uphill battle to keep up with, but with serilog being pretty ubiquitous and the use of a command interceptors I think this one provides value to a decent amount of users.


patriksvensson (approved):
LGTM!

PR #288, opened 2021-03-06
Adds overloads for MarkUp methods without args

### LINKED ISSUE #287: Maybe bug with signle curly braces and EscapeMarkup
**Information**
 - OS: Windows 10, 10.0.21313
 - Version: 0.37.0
 - Terminal: Windows Terminal, conhost

**Describe the bug**
Using curly braces symbol in the source text causes the "Input string was not in a correct format" exception when calling the MarkupLine function.
And function EscapeMarkup is not escape curly braces symbol.

**To Reproduce**
 ```
 var escapedText = "{".EscapeMarkup();
 MarkupLine($"[red bold]{escapedText}[/]");
 ```

**Expected behavior**
Success rendering

**Additional context**
I understand that `{` symbol is used in formatting when placing arguments passed to the `MarkupLine` function.
And probably this is solved by the usual extension for a string that will safely wrap symbol `{` into `{{`.
But still, I would like to know if this exception is correct?

phil-scott-78:
These methods don't require a `string.Format` call so we'll directly call the Render method without a call to `string.Format`. 

Added bonus of a couple fewer allocations too.

Closes #287

### DIFF
--- a/src/Spectre.Console.Tests/Unit/MarkupTests.cs
+++ b/src/Spectre.Console.Tests/Unit/MarkupTests.cs
@@ -71,5 +71,19 @@ namespace Spectre.Console.Tests.Unit
             // Then
             console.Output.ShouldBe(output);
         }
+
+        [Fact]
+        public void Should_not_fail_with_brackets_on_calls_without_args()
+        {
+            // Given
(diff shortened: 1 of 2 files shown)


patriksvensson:
@phil-scott-78 There seem to be some check failures in this PR. Not sure why the build doesn't fail.

patriksvensson (approved):

PR #297, opened 2021-03-13
Hide examples of hidden commands

### LINKED ISSUE #290: Examples of hidden commands are shown in the help text
**Information**
 - OS: All three: Windows, Linux, MacOS
 - Version: 0.37.1-preview.0.36
 - Terminal: Windows Terminal

**Describe the bug**
If you add an example for a command that is hidden, the example shows in the root command's help text.

**To Reproduce**

```c#
config
    .AddCommand<MyCommand>("test")
    .WithDescription("This is a test command")
    .WithExample([| "test", "--some-arg", "123" |])
    .IsHidden();
```

**Expected behavior**
Examples of hidden commands should only show up in the help text when explicitly asking for *that* command's help text. For commands that are a root of that command it should not show the example, including the default command.

**Additional context**
Note: The command is actually hidden correctly, it's just the example that shows up in the help text.

bddckr:
Fixes #290.


patriksvensson (approved):
LGTM!

PR #298, opened 2021-03-13
Add support for exclusive mode

patriksvensson:
Added `IExclusivityMode` that can be used to prevent two things that want exclusivity to run simultaneously, i.e., a prompt and a status updater. I'm not saying this should be there forever, but at least until we've figured out how to handle cases like this properly.

Not sure about the name for this and the error messages, so consider this a draft that needs some input 😅

### DIFF
--- a/docs/input/prompts/multiselection.md
+++ b/docs/input/prompts/multiselection.md
@@ -7,6 +7,11 @@ one or many items from a provided list.
 
 <img src="../assets/images/multiselection.gif" style="width: 100%;" />
 
+<div class="alert alert-warning" role="alert" style="margin-top:20px;">
+  <i class="fas fa-exclamation-triangle icon-web"></i> The use of prompts 
+  insides status or progress displays is not supported.
+</div>
+
 # Usage
 
 ```csharp
--- a/src/Spectre.Console/AnsiConsoleSettings.cs
+++ b/src/Spectre.Console/AnsiConsoleSettings.cs
@@ -31,6 +31,11 @@ namespace Spectre.Console
         public InteractionSupport Interactive { get; set; }
 
         /// <summary>
+        /// Gets or sets the exclusivity mode.
+        /// </summary>
+        public IExclusivityMode? ExclusivityMode { get; set; }
+
+        /// <summary>
         /// Gets or sets the profile enrichments settings.
         /// </summary>
         public ProfileEnrichment Enrichment { get; set; }
--- a/src/Spectre.Console/IExclusivityMode.cs
+++ b/src/Spectre.Console/IExclusivityMode.cs
@@ -1,0 +1,27 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Spectre.Console
+{
+    /// <summary>
+    /// Represents an exclusivity mode.
+    /// </summary>
+    public interface IExclusivityMode
+    {
+        /// <summary>
+        /// Runs the specified function in exclusive mode.
+        /// </summary>
+        /// <typeparam name="T">The result type.</typeparam>
+        /// <param name="func">The func to run in exclusive mode.</param>
+        /// <returns>The result of the function.</returns>
+        T Run<T>(Func<T> func);
+
+        /// <summary>
+        /// Runs the specified function in exclusive mode asynchronously.
+        /// </summary>
+        /// <typeparam name="T">The result type.</typeparam>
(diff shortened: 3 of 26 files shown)


src/Spectre.Console/IExclusivityMode.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Spectre.Console
+{
+    /// <summary>
+    /// Represents an exclusivity mode strategy.
+    /// </summary>
+    public interface IExclusivityMode
```
patriksvensson, line 9: Not sure about this name 😅

src/Spectre.Console/Internal/DefaultExclusivityMode.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            _semaphore = new SemaphoreSlim(1, 1);
+        }
+
+        public T Run<T>(Func<T> func)
+        {
+            // Try aquiring the interactive semaphore
+            if (!_semaphore.Wait(0))
+            {
+                throw new InvalidOperationException(
```
patriksvensson, line 21: Not sure if this error message is good or if it can be expressed in some other way to make it obvious to the user what's happening.

phil-scott-78: FWIW, I think the original message would be "e.g.  a prompt and a progress display", if my latin is correct.

I think something like "Trying to run one or more interactive functions concurrently. Operations with dynamic displays (e.g. a prompt and a progress display) cannot be running at the same time." would be good, specifically giving these types of operations some sort of name that we can reference in the error message that also matches what is in the documentation.

phil-scott-78:
Everything looks good to me, code wise at least.

I think one thing that would need to happen with these release would be a quick reordering of the documentation. I'd move Progress and Status together under one menu group named something like "Live Displays" or "Interactive Command". Group these or . Under the `index.md` for that group I'd include a warning that live displays like the progress and status cannot be ran simlutaneously with each other or with text prompts. Likewise I'd include a warning under the `index.md` for Prompts too.

Heck, with documentation in place I'd be tempted to include the url to the docs in the exception we are throwing to the warning itself.

patriksvensson:
@phil-scott-78 Made some updates according to your feedback.

docs/input/prompts/multiselection.md
```diff
@@ -7,6 +7,11 @@ one or many items from a provided list.
 
 <img src="../assets/images/multiselection.gif" style="width: 100%;" />
 
+<div class="alert alert-warning" role="alert" style="margin-top:20px;">
+  <i class="fas fa-exclamation-triangle icon-web"></i> Using prompts inside 
+  status or progress displays, are not supported.
```
phil-scott-78, line 12: I'm like 95% sure this should be "Using prompts insides status or progress displays is not supported" or maybe even better "The use of prompts insides status or progress displays is not supported"

phil-scott-78:
added an absolutely trivial grammar suggestion, but I wanted to pull down and try and break this tomorrow before merging. Everything looks solid but with it effecting people immediately I want to give a test go.

src/Spectre.Console/AnsiConsoleSettings.cs
```diff
@@ -30,6 +30,11 @@ public sealed class AnsiConsoleSettings
         /// </summary>
         public InteractionSupport Interactive { get; set; }
 
+        /// <summary>
+        /// Gets or sets the exclusivity lock.
+        /// </summary>
+        public IExclusivityMode? ExclusivityLock { get; set; }
```
bddckr, line 36: I'm nitpicking here, sorry about that 😅

Is it a "mode" or a "lock"? Should this be consistent (including the naming in the tests)?

patriksvensson: @bddckr It's "Mode". Thanks!

phil-scott-78: I'm also liking keeping the interface named Mode.

In this case the implementation is a lock, but with fake implementation there is no lock. And heck, who knows - someone much more clever than me might even be able to cook up with a way to support multiple progress bars with their own custom exclusivity mode in the future.

patriksvensson:
@phil-scott-78 Should be ready for review

PR #299, opened 2021-03-14
Fix Windows ANSI support detection with redirected stderr

bastianeicher:
Currently the Windows ANSI support detection does not take into consideration whether the `AnsiConsole` was created with stdout or stderr as the target.

Assuming:
```csharp
var stderr = AnsiConsole.Create(new AnsiConsoleSettings {Out = Console.Error});
```

If you redirect stderr (e.g. `myapp 2> errors.txt`), `stderr.Profile.Capabilities.Ansi` will still be `true` even though it should be `false`.

Conversely, if you redirect stdout but not stderr (e.g. `myapp > outputs.txt`), `stderr.Profile.Capabilities.Ansi` will still be `false` even though it should be `true`.

I added a new parameter `bool stdError` to `AnsiDetector.Detect` to address this.

### DIFF
--- a/src/Spectre.Console/AnsiConsoleFactory.cs
+++ b/src/Spectre.Console/AnsiConsoleFactory.cs
@@ -68,7 +68,7 @@ namespace Spectre.Console
 
             if (settings.Ansi == AnsiSupport.Detect)
             {
-                (supportsAnsi, legacyConsole) = AnsiDetector.Detect(true);
+                (supportsAnsi, legacyConsole) = AnsiDetector.Detect(buffer.IsStandardError(), true);
 
                 // Check whether or not this is a legacy console from the existing instance (if any).
                 // We need to do this because once we upgrade the console to support ENABLE_VIRTUAL_TERMINAL_PROCESSING
@@ -93,7 +93,7 @@ namespace Spectre.Console
                         else
                         {
                             // Try detecting whether or not this
-                            (_, legacyConsole) = AnsiDetector.Detect(false);
+                            (_, legacyConsole) = AnsiDetector.Detect(buffer.IsStandardError(), false);
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


patriksvensson:
Not sure I agree here 100%. Even if console is a tty, ansi capability is just an indication of whether the current environment (console/terminal) support VT sequences or not.

Not at a computer atm, but will read theough this again when I'm at a computer.

bastianeicher:
Consider the following code snippet:

```csharp
var errorConsole = AnsiConsole.Create(new AnsiConsoleSettings {Out = Console.Error});
errorConsole.WriteLine("Hello world", new Style(Color.Red));
```

When running this with `dotnet run` the output looks as expected:
ansi

However, when running `dotnet run > output.txt` the output looks like this:
no-ansi

So Spectre.Console disabled ANSI colors for stderr even though only stdout was redirected. Similar behavior can be observed with things like `Progress` falling back to it's fallback renderer.

My thinking was, that `Capabilities.Ansi` indicates whether ANSI escape code should be used for the `AnsiConsole` in question.
When output is not redirected, this should be true if the current console/terminal supports it.
When output is redirected this should always be false, because escape codes are of no use in text files or when piped into other processes.

patriksvensson:
@bastianeicher Not sure I follow tbh.

Why would redirecting stderr affect things that write to stdout? Capabilities are only valid for redirected stdout as it is now. Not sure that we would want to tell Spectre.Console that stdout doesn't support ANSI just because stderr doesn't?

It's late and I'm tired so feel free to correct me here.

### Redirecting stderr to file:

image

### Redirecting stderr and stdout to file:

image

bastianeicher:
Sorry, I was being a bit unclear.

Currently, on Windows, *redirecting stdout affects things that write to stderr*, even though it should not. That's what my sample above illustrates. This problem does not occur on Linux.

With this PR that problem goes away.

patriksvensson (approved):

patriksvensson:
@bastianeicher Ok, I think I understand 👍🏻

patriksvensson:

Issue #296, opened 2021-03-12
SelectionPrompt: Support PgUp, PgDown, Home, and End keys


patriksvensson:
**Is your feature request related to a problem? Please describe.**
Would be great if we supported `PgUp`, `PgDown`, `Home`, and `End` keys for `SelectionPrompt`

**Additional context**
I can take a stab at this tonight

chucker:
Shouldn't this work as of 0.46.0? Scrolling with Up/Down works, but paging with PgUp/PgDown doesn't seem to.

Issue #229, opened 2021-01-14
Allow custom texts in prompts


angelobreuer:
**Is your feature request related to a problem? Please describe.**
It would be useful to customize texts for prompts.

**Why is this needed?**
I need custom texts in prompts for localization purposes, but I think it would be beneficial to allow custom texts.

**Describe the solution you'd like**
A solution to customize texts in prompts.

**Additional context**

```cs
var prompt = new MultiSelectionPrompt<string>()
    .ShowMoreOptions(text: "Press [Up] or [Down]-Arrow to show more.")
    ...
```

patriksvensson:
@angelobreuer I absolutely agree 👍🏻

Issue #221, opened 2021-01-12
Allow SelectionPrompt choices to have markup


Nikey646:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
To be able to use markup in choices for the `SelectionPrompt<T>` class.

**Describe alternatives you've considered**
N/A

**Additional context**

angelobreuer:
I would also need this (+1).
I am available to grab this issue if this would be a candidate.

PR #300, opened 2021-03-14
Various enhancements to selection prompts

patriksvensson:
This PR adds some requested features for selection prompts.

* Support Home, End, PgUp, and PgDown in prompts (#296)
* Allow markup in selection prompts (#221)
* Allow custom instructions for prompts (#229)


phil-scott-78 (approved):
Ran through the example. New messaging really helps out. First time I ran the same I didn't realize I could scroll up and down.

Everything else looks solid.

PR #308, opened 2021-03-20
Add support for alternate screen buffers

### LINKED ISSUE #250: Support alternate screen buffer?
[Be neat if main and alt buffers were supported](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#alternate-screen-buffer). I want my app to do everything in one buffer then switch back to the main one on exiting and restore the original screen buffer

Maybe something like 

```csharp
AnsiConsole.Console.SetBuffer(Buffer.Main)
AnsiConsole.Console.SetBuffer(Buffer.Alternate)
```

```
AnsiConsole.Console.Write(Segment.Control("\u001b[?1049h")); // alternate
// do stuff
AnsiConsole.Console.Write(Segment.Control("\u001b[?1049l")); // main
```

**Additional context**
This might be cruising for a bruising. Just playing with it by sending in the escape codes by hand screws with things it seems. E.g. tables for some reason get rendered on both on my test app.

patriksvensson:
This is a draft for now. Need to think about naming a bit more to keep things consistent with terminal terminology.

Closes #250


src/Spectre.Console/AlternateBuffer.cs
```diff
@@ -0,0 +1,42 @@
...
+using Spectre.Console.Rendering;
+
+namespace Spectre.Console
+{
+    /// <summary>
+    /// Represents an alternative screen.
+    /// Once created, all output the console will be written
+    /// to the alternate screen until disposed.
+    /// </summary>
+    public sealed class Screen : IDisposable
```
patriksvensson, line 11: Perhaps rename this to `ScreenBuffer`?

src/Spectre.Console/Capabilities.cs
```diff
@@ -15,6 +15,12 @@ public sealed class Capabilities
         /// </summary>
         public bool Ansi { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether or not the
+        /// terminal supports alternating buffers.
+        /// </summary>
+        public bool AlternateBuffer { get; set; }
```
patriksvensson, line 22: Need to figure out how we can detect this properly for different terminals. Right now I'm just enabling it for Windows Terminal.

src/Spectre.Console/AlternateBuffer.cs
phil-scott-78: I was thinking about how I might use this class. I might be tempted to write some code to run my entire app in the alternate screen buffer (ala vi) like

``` csharp
void Main() {
    using var screen = new Screen(Console);
    // my entire app in the alt buffer
}
```

Maybe a name that fits that usage like `AlternateScreenSwitch`?

patriksvensson: @phil-scott-78 Not really a fan of the `Switch` postfix to be honest, but perhaps `AlternateScreen`?

Entomy:
Per @phil-scott-78's request I took a bit of a dive into the sources to compare our approaches. Looks like we're nearly doing the same thing.

I used a `ref struct` just because it was more light weight. There's no functional reason for this. The type itself doesn't really do anything, it just exists for the purposes of exploiting how disposable types work, so I figured the most light weight option would suffice.

From the user standpoint, activating the buffer was as follows:

~~~~csharp
using ScreenBuffer Screen = Console.ScreenBuffer;
~~~~

Looks like @phil-scott-78 has a similar idea.

I opted to enable this through a property, as follows:

~~~~csharp
/// <summary>
/// Gets the alternate screen buffer, and actives it.
/// </summary>
public static ScreenBuffer ScreenBuffer {
	get {
		ConsoleComponents.StateManager.UseAlternateScreenBuffer();
		return new ScreenBuffer();
	}
}
~~~~

The reason for this was easier discover of the alt buffer mechanism, since it would show up under intellisense for `Console.*`. It didn't occur to me until reviewing this, but it seems like we both have an unintended consequence of our designs. It's probably insignificant, but something to consider non-the-less. We're both activating the alt buffer on access/construction and deactivating it on destruction. What happens when a more complicated app activates the alt buffer in two locations? Our "proxy" has two separate lifetimes, and one will be destructed while the other is still in use, deactivating the alt buffer while another still wants it. In practice this sounds like a code smell, and I doubt it would be run into very often, but it's something to keep in mind. You have far more users, and maybe a better idea if this is an issue for any of them.

Otherwise, ignoring architectural differences, we're doing almost the exact same thing. At least from my own uses, the approach worked well. Doesn't get any better than a single line of code to activate, that cleans itself up.

patriksvensson:
Closing this for now until I've reworked it a bit.

PR #321, opened 2021-03-24
Change `IAnsiConsole` to render `IRenderable` instead of `IEnumerable<Segment>`

patriksvensson:
This makes it possible for encoders to output better representation of the actual objects instead of working with chopped up segments.

* `IAnsiConsole.Write` now takes an IRenderable instead of segments
* Calculating cell width does no longer require a render context
* Removed `RenderContext.LegacyConsole`
* Removed `RenderContext.Encoding`
* Added `Capabilities.Unicode`


src/Spectre.Console/Internal/Backends/Ansi/AnsiConsoleCursor.cs
```diff
@@ -16,11 +15,11 @@ public void Show(bool show)
         {
             if (show)
             {
-                _backend.Render(new[] { Segment.Control("\u001b[?25h") });
+                _backend.Write(new ControlSequence("\u001b[?25h"));
```
phil-scott-78, line 18: a thought for the future but not this PR - be nice if `ControlSequence` had some of these values as constants, or in the case of the ones like "move up x characters" there is a function that generates the value. To the untrained eye `\u001b[?25h` not only looks like gibberish, but is pretty difficult to google.

phil-scott-78 (approved):

phil-scott-78:
Finally got a chance to sit down and play with the update last night. No surprises on my end, and the code looks good to me.

But with it being such a big change I wanted to make sure we both turned the key at the same on this launch before I merge it so just let me know when you are all set.

src/Spectre.Console/Internal/Backends/Ansi/AnsiConsoleCursor.cs
patriksvensson: Absolutely! I will send another PR for this!

patriksvensson:
@phil-scott-78 Actually noticed a bug so might be a good idea to wait merging this for the time being

PR #334, opened 2021-04-02
Add support for indeterminate progress

### LINKED ISSUE #329: Add support for indeterminate progress
Show an indication that something is happening when we don't know how long an operation will take. This will show a cyclic animation without indicating a specific amount of progress.

```csharp
AnsiConsole.Progress()
    .Start(ctx => 
    {
        // Create a task in a indeterminate state
        var task = ctx.AddTask("[green]Folding space[/]", autoStart: false);

	// Omitted...
    });
```

### LINKED ISSUE #331: Progress bars does not render correctly in NO_COLOR
When the current terminal is set to `NO_COLOR`, progress bars aren't rendered correctly.

image

patriksvensson:
This commit also changes the behavior of ProgressContext.IsFinished.
Only tasks that have been started will be taken into consideration, and not indeterminate tasks.

Closes #329
Closes #331

### DIFF
--- a/examples/Console/Progress/Program.cs
+++ b/examples/Console/Progress/Program.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using Spectre.Console;
 
@@ -25,8 +26,12 @@ namespace ProgressExample
                 .Start(ctx =>
                 {
                     var random = new Random(DateTime.Now.Millisecond);
-                    var tasks = CreateTasks(ctx, random);
 
+                    // Create some tasks
+                    var tasks = CreateTasks(ctx, random);
+                    var warpTask = ctx.AddTask("Going to warp", autoStart: false).IsIndeterminate();
+
+                    // Wait for all tasks (except the indeterminate one) to complete
                     while (!ctx.IsFinished)
                     {
                         // Increment progress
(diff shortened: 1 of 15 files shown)


phil-scott-78:
Code looks solid, and the animation looks great. 

I think this is what we talked about earlier, but while I'm thinking about it it seems to me a task has 4 states

1. Not started
2. Started and we know an amount of work to be done (determinate)
3. Started but we don't know how much work is left to be done (indeterminate)
4. Finished

I'd think a task being indeterminate should be opt in like 

```csharp
var warpTask = ctx.AddTask("Going to warp", autoStart: false).IsIndeterminate();
```

And in that example no progress bar would be displayed (and other columns would follow suit). Then once it is started, then it moves to state 3 rendering the fancy graphics until we reach step 4 in which case we render it as finished.

I think moving from step 2 to 3 and back again is also a valid use case too (e.g. like an internet connection disconnecting during a download going to indeterminate and then back to regular when the connection resumes).

Sorry if this duplicates our chat earlier, but someone else might also have feedback so I wanted to get it all public too.

phil-scott-78:
Code looks solid, and the animation looks great. 

I think this is what we talked about earlier, but while I'm thinking about it it seems to me a task has 4 states

1. Not started
2. Started and we know an amount of work to be done (determinate)
3. Started but we don't know how much work is left to be done (indeterminate)
4. Finished

I'd think a task being indeterminate should be opt in like 

```csharp
var warpTask = ctx.AddTask("Going to warp", autoStart: false).IsIndetm;

**Comment by patriksvensson** [comment]
@phil-scott-78 Great that you got that down.

Question is, should there be an explicit way of turning off indeterminate state, like `task.NotIndeterminate()`? I think we will need that to transition between `2` and `3` for example.

**Comment by patriksvensson** [comment]
@phil-scott-78 The API is going to looks something like this then:

```csharp
// Create some tasks
var tasks = CreateTasks(ctx);
var warpTask = ctx.AddTask("Going to warp", autoStart: false).IsIndeterminate();

// Wait for all tasks (except the indeterminate one) to complete
while (!ctx.IsFinished)
{
    // Omitted, but will increment all started tasks
    // all of them are completed.
}

// Now start the "warp" task
warpTask.StartTask();
warpTask.IsIndeterminate(false);
while (!ctx.IsFinished)
{
    warpTask.Increment(12 * random.NextDouble());

    // Simulate some delay
    Thread.Sleep(100);
}
```

patriksvensson:
@phil-scott-78 Updated the PR with your suggestions.

phil-scott-78 (approved):
Everything looks great. Demo looks especially good.

PR #335, opened 2021-04-02
Rename tree to root so the example compiles.

mbdev87:
The 'root' variable does not exist in current context. Either should call Render(tree) or have it renamed to root. I've chosen root as it's more consistent with the examples later on.


patriksvensson (approved):

patriksvensson:
Merged! Much appreciated! 👍

PR #336, opened 2021-04-03
Replaces emoji regex with ReadOnlySpan implementation

phil-scott-78:
The RegEx runtime perf was never anything noticeable - it was the startup time that was eating over a third of time during initialization. 
 This shaves 200ms off the startup time.

image

I tried to add as many tests of combinations as possible and the sample apps and things seem to be working just as originally, but if you could hammer on it and try to break it. 

Double check my `ifdef` for netstandard too. It's a little ugly, but should keep netstandard 2.0 support chugging along.

### DIFF
--- a/src/Spectre.Console.Tests/Unit/EmojiTests.cs
+++ b/src/Spectre.Console.Tests/Unit/EmojiTests.cs
@@ -41,5 +41,56 @@ namespace Spectre.Console.Tests.Unit
                 result.ShouldBe("Hello 🌍!");
             }
         }
+
+        public sealed class Parsing
+        {
+            [Theory]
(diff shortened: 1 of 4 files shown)


patriksvensson (changes requested):
Other than some minor stuff, this looks very good. Great work!

phil-scott-78:
Great feedback and I appreciate it! I added an extension method but named it `AppendSpan` just so there wasn't a conflict with newer frameworks which already have the `Append`. 

Amended the commit and pushed the changes up.

patriksvensson (approved):

patriksvensson:
Merged 👍

PR #339, opened 2021-04-04
Fix OS version detection

shana:
Hello! 👋

This is a quick patch to fix color detection. From the comparison `build >= 15063`, I assume this was meant to use the `Build` part of the OS version, and not the `Minor` part (since Windows 10 versions have the format 10.0.XXXXX).


phil-scott-78 (approved):

phil-scott-78:
You are absolutely right. Great catch by you, big miss by me. I'll get this merged and a new nuget package. will be pushed shortly

PR #340, opened 2021-04-04
Character cells count perf

phil-scott-78:
I noticed that we spend a lot of time calculating widths of strings and their rune size. E.g. to render the progress demo, it is over 1,000,000 measurements. This is a proposal to cache those values for quick access.

Because there are a limited number of runes this adds a static array to cache them that we can then just go right to their memory spot without needing to hit WcWidth for these values. Being able to quickly grab these values drops rendering time significantly, especially if we are rendering something repeatably like a progress bar.

# Simple example using the tables demo.

Before
image

After
image

# More Complex

Here's one rendering a tree. Nearly an entire second was spent counting characters

Before
image

After
image

# Even More Complex.

Running the progress demo shows even more improvements. shaves 3 seconds off of the millions of calculations we do.

Before
image

After
image

Downside, of course, is we are eating up 65kb to store this data but that's a pretty fair tradeoff. Might make sense to looking at reducing the calls to the method to begin long term.

### DIFF
--- a/src/Spectre.Console/Internal/Cell.cs
+++ b/src/Spectre.Console/Internal/Cell.cs
@@ -4,11 +4,14 @@ namespace Spectre.Console
 {
     internal static class Cell
     {
+        private static readonly int?[] _runeWidthCache = new int?[char.MaxValue];
+
         public static int GetCellLength(string text)
         {
             var sum = 0;
-            foreach (var rune in text)
+            for (var index = 0; index < text.Length; index++)
             {
+                var rune = text[index];
                 sum += GetCellLength(rune);
             }
 
@@ -28,7 +31,7 @@ namespace Spectre.Console
                 return 1;
             }
 
-            return UnicodeCalculator.GetWidth(rune);
+            return _runeWidthCache[rune] ??= UnicodeCalculator.GetWidth(rune);
         }
     }
 }


phil-scott-78:
man did I screw something up rebasing my local against the main branch. ignore those commits. I'll get that straightened up once the kiddo takes a nap.

phil-scott-78:
well things went totally squirelly. Disregard for now, I've done terribly dumb.

phil-scott-78:
And on the third `git push --force`, the build is risen.

phil-scott-78:
I'm going to try a few more benchmarks, so hold off on a merge for now.

phil-scott-78:
Ok, I feel pretty good about the speed. Used BenchmarkDotNet to calculate the cell length of 1000 random strings company names from Bogus. Here are the results.

|      Method |      Mean |    Error |   StdDev |
|------------ |----------:|---------:|---------:|
|    Original | 907.34 us | 6.609 us | 5.858 us |
|   WithCache |  60.90 us | 0.555 us | 0.519 us |

Of course, this means that the original method is averaging less than one-microsecond. But with us doing potentially millions of them it's worth it. FWIW, I also tested out using 

```csharp
StringInfo.ParseCombiningCharacters(text).Length
```

but not only did it break the tests with things like chinese characters, but it was also 10x slower than the cached character implementation.

src/Spectre.Console/Internal/Cell.cs
```diff
@@ -28,7 +31,7 @@ public static int GetCellLength(char rune)
                 return 1;
             }
 
-            return UnicodeCalculator.GetWidth(rune);
-        }
+            return _runeWidthCache[rune] ??= UnicodeCalculator.GetWidth(rune);
+ }
```
patriksvensson, line 35: @phil-scott-78 There seems to be something wrong with the indentation here.

patriksvensson (changes requested):
Looks good to me. Just a minor nitpicking thing that needs to be fixed 😅

phil-scott-78:
good catch. fixed that up and ammended the commit

patriksvensson (approved):
LGTM!

PR #341, opened 2021-04-04
Fix AnsiConsoleBackend.Clear

### LINKED ISSUE #337: IAnsiConsole.Clear() does not clear the scroll buffer
**Information**
 - Version: Most recent master: 6121203fee21c2b7f73bb8b1729147e3dce5bdb9

**Describe the bug**
The `IAnsiConsole.Clear()` command does not clear the scroll buffer.

**To Reproduce**
```csharp
using Spectre.Console;
using System;

for (int i = 0; i < 100; i++)
{
    AnsiConsole.WriteLine($"Line {i}");
}

AnsiConsole.Console.Clear(true);
```

**Expected behavior**
I expect the `.Clear()` method to clear everything like `System.Console.Clear()` does.

**Screenshots**
Here is a screenshot after the `Clear()` call, with the example provided above
image

**Solution**
I took a look at the source and the wikipedia article for ANSI escape codes and it looks like the clear method isn't implemented correctly.

At the moment is looks like that:
```csharp
// Spectre.Console/Internal/Backends/Ansi/AnsiConsoleBackend.cs
public void Clear(bool home)
{
    Write(new ControlSequence(ED(2)));
    if (home)
    {
        Write(new ControlSequence(CUP(0, 0)));
    }
}
```

According to Wikipedia ED(2) only clears the screen buffer (which the current behaviour does). I also noticed that CUP(0, 0) is wrong because CUP values are 1-based and CUP(1, 1) would be top left.

IMO the method should be changed to
 - ED(3) - _Which would clear the entire buffer like `System.Console.Clear` does_ and
 - CUP(1, 1)

wischi-chr:
## Add an ED3 command to clear the scroll buffer.

This command is not in the original ANSI terminal control sequences but
was added later to XTerm to account for the fact that most modern
terminals have a scroll-buffer.

https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h4-Functions-using-CSI-_-ordered-by-the-final-character-lparen-s-rparen:CSI-Ps-J.1C8A

This command is now awailable in basically
all terminals with scroll-buffers

## Change CUP coordinates to 1, 1

The coordinates 0, 0 also worked, but strictly speaking they are outside
the available screen area and are capped to 1, 1 anyways.

This commit fixes #337

### DIFF
--- a/src/Spectre.Console.Tests/Unit/AnsiConsoleTests.cs
+++ b/src/Spectre.Console.Tests/Unit/AnsiConsoleTests.cs
@@ -8,8 +8,8 @@ namespace Spectre.Console.Tests.Unit
     public partial class AnsiConsoleTests
     {
         [Theory]
-        [InlineData(false, "Hello[2JWorld")]
-        [InlineData(true, "Hello[2J[0;0HWorld")]
+        [InlineData(false, "Hello[2J[3JWorld")]
+        [InlineData(true, "Hello[2J[3J[1;1HWorld")]
         public void Should_Clear_Screen(bool home, string expected)
         {
             // Given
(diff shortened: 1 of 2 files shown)


patriksvensson (approved):
LGTM! 👍

patriksvensson:

PR #346, opened 2021-04-05
Falls back to null for settings constructor

phil-scott-78:
To be honest, I'm not quite sure if this is the right approach or if I'm even detecting valid null properties. What I'm trying to do is have my settings configured like so

```csharp
public sealed class Settings : CommandSettings
{
    public Settings(string[]? command, bool? detailed)
    {
        Command = command;
        Detailed = detailed;
    }

    [CommandArgument(0, "[command]")]
    public string[]? Command { get; }

    [Description("Include detailed information about the commands.")]
    [CommandOption("-d|--detailed")]
    public bool? Detailed { get; }
}
```

In a situation where neither params are passed I'd want the constructor called with `.ctor(null, null)`. As it stands now it tries to resolve `string[]` and fails.


patriksvensson (approved):
LGTM!

PR #348, opened 2021-04-05
Remove redundant Style constructor.

wischi-chr:
Remove the redunant constructor. It doesn't even have aesthetic advantages because it's a private constructor.

### DIFF
--- a/src/Spectre.Console/Style.cs
+++ b/src/Spectre.Console/Style.cs
@@ -34,11 +34,6 @@ namespace Spectre.Console
         /// </summary>
         public static Style Plain { get; } = new Style();
 
-        private Style()
-            : this(null, null, null, null)
-        {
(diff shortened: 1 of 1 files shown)


phil-scott-78:
This actually isn't redundant, if you check right above the constructor, you'll see 

```csharp
public static Style Plain { get; } = new Style();
```

Since that's within this class the private constructor gets called resulting in a true plain style with no foreground, background or text decoration set like a public call to a the `ctor` which would get a default foreground, background, and text decoration.

patriksvensson:
It's probably a forgotten remnant of old times that survived a refactor 😃

phil-scott-78 (approved):

phil-scott-78:
Got it merged. Thanks for taking the time to submit.

PR #349, opened 2021-04-05
Move GetSegments to AnsiConsoleExtensions

wischi-chr:
Swap the parameters to make the "active" type (in this case the console)
the type that is extended and the "passive" renderable the parameter.

There are some indications that the parameters are currently reversed.
For example in `FakeConsole.Write` the object passes itself into a method of another type.

```csharp
foreach (var segment in renderable.GetSegments(this))
```

In my experience such cases are not very common. IMHO `this.GetSegments(renderable)` would be more natural.

Another location that might indicate reversed parameters is the implementation itself. In the method the console is used on the left-hand side and the renderable is passed as a parameter to the console's pipeline.

```csharp
var renderables = console.Pipeline.Process(context, new[] { renderable });
```

One might argue that a few lines later the renderable is the "active" instance. That's true, but even in that case the console isn't passed as a parameter to the renderable. The renderable just needs to know the width. I restructured that part to make that more clear _(and in the process removed the private method and the intermediate `List<Segment>`)_.


phil-scott-78:
I kicked off the build job, issue was with nuget restore not the code itself.

As to the commit itself, I can see the argument you are making. I'm not a huge fan of having more methods hanging off `AnsiConsole` though. That's the first-class object that people will be using for discovery when using the library. That experience is important too. Heck, I'd be tempted to drop the extension method altogether before adding more extensions to `AnsiConsole`, especially one that doesn't involve direct I/O. @patriksvensson can provide some historical guidance too that I might be overlooking.

Regardless, if we went this route, I'm not sure deleting the current extension method would be the move. Probably would need to leave it as `Obsolete` for a few releases at least I'd think.

patriksvensson:
@wischi-chr I appreciate your work with this, but I think I agree with @phil-scott-78 here. I would prefer to keep these kinds of things away from `IAnsiConsole` for the reasons @phil-scott-78 brought up. It's not perfect, but it's a work in progress. At some point we there will be changes internally to these kinds of things, and perhaps we can revisit this then.

wischi-chr:
Yeah sure, no problem. Feel free to close the PR, no hard feelings about that.

phil-scott-78:
I'm going to close this PR. But I'd like to echo @patriksvensson thanks. For each release I've been trying to give a small thanks to contributors when I mention it on twitter. If you have a twitter handle let me know so I can thank you for your help the with the issues and fixes you've provided that will be in the next release!

wischi-chr:
I'm not really a twitter user, but thank you for the offer.

Issue #314, opened 2021-03-23
Add ability to mark ProgressTask as failed


Tyrrrz:
**Is your feature request related to a problem? Please describe.**
When performing multiple parallel tasks, it's likely that some of them may fail without stopping the entire process. It would be nice if there was a way to signal to the user that a task has failed.

**Describe the solution you'd like**
Currently, `ProgressTask` exposes a method called `StopTask()`. The least intrusive solution I see is to add an overload called `StopTask(bool isError)`.

If the task is stopped with an error, we can apply some kind of `ErrorStyle`, which would make the progress bar appear red. Additionally, we can make it so a failed task displays `X` somewhere, for example in the `PercentageColumn`.

**Describe alternatives you've considered**
I considered writing something to the console myself, but it breaks the progress reporting.

**Additional context**

Here's how it currently looks:

![image](https://user-images.githubusercontent.com/1935960/112215867-8a7e1a80-8c29-11eb-8315-92f13de1938f.png)

One of the tasks has failed, but its progress remains at 0 with no visual indication that it crashed. As a user, it's not clear whether it's just stuck at 0% or it failed.

redoz:
@patriksvensson would you take a PR for this? I was thinking of adding `bool Failed { get; set;}` to the `ProgressTask` and an additional property (`Style FailedStyle {get;set;} = Color.Red`?) to the `ProgressBarColumn`

PR #354, opened 2021-04-08
Add fail task

GabrielNarcisi:
fix Issue #314 by adding IsFailed to ProgressTask


patriksvensson:
@GabrielNarcisi There are some build problems with your PR which need to be fixed before we can review it.

PR #371, opened 2021-04-13
Adds default command to "explain" and hides hidden commands

phil-scott-78:
* Default command wasn't included in the list
* Hides the hidden commands by default with a flag to override the behavior (`--hidden`)


patriksvensson (approved):
LGTM!

PR #372, opened 2021-04-13
Add output abstraction and reorganize profile

### LINKED ISSUE #343: Move ColorSystem into Capabilities
It's not a problem with the behavior but with how the `Profile` is currently structured.

The `ColorSystem` property should be moved to the `Capabilities` class because it really represents one of the capabilities of the console.

The class `RenderContext` makes this even more evident. Because `ColorSystem` is not inside `Capabilities` / `IReadOnlyCapabilities` it results in a _(IMHO)_ strange egde case where it has to be passed in as a separate constructor parameter and stored in a separate property inside `RenderContext`.

The alternative would be to leave it as it is and I have to live with the fact that it's not implemented how I would do it 😅

### LINKED ISSUE #359: A way to to use System.Console.WindowWidth for redirected TextWriters.
In the current implementation if one uses a custom TextWriter/Output Stream the `Tty` property returns true and this automatically leads to a situation where `GetHeight()` and `GetWidth()` return some default constants that can't be controlled.

A user of the library can't even prevent that because Profile and Capabilities are both sealed.

I currently use a custom TextWriter that later down the line ultimately writes to the real console out TextWriter, but the way Tty is implemented breaks the width and height detection an causes the output to wrap after 80 characters.

I'm pretty open to ideas. The end result should be a situation where I can use a custom TextWriter and can somehow hook GetWidth and GetHeight. Manually setting width and height is not a good option because for a real console (like in my case) there are no good options to get a resize event so I would have to constantly poll `System.Console.WindowWidth` and set the width for the profile.

Things I would try/consider:

- A way to overwrite the Tty flag (introduce a setter with a nullable backing field, like width and height in `Profile`)
_Probably the simplest solution?_
- Unseal `Capabilities` so users can change the tty behaviour
- Unseal `Profile` so it can be derived from
- Introduce an `IProfile` interface so `IAnsiConsole` implementations can provide a different profile implementation

Polling `System.Console.WindowWidth` and set the width for the profile (nothing I would _seriously_ consider) or somehow fix it in a custom fork. Another possibility is to set `Profile.Width = int.MaxValue;`  and let the terminal do the line breaks, but this of course breaks almost all widgets.

### LINKED ISSUE #369: Capabilities.Tty is confusingly named
`Capabilities.Tty` is documented as "Gets a value indicating whether or not console output has been redirected."

This sounds like the opposite of what a `bool Tty` property would mean. Where I'd expect true to mean "this console is a TTY handle" or possibly "this console is a TTY or PTY handle". 

I'd suggest either renaming the property to `IsRedirected` or flipping the current logic around it so it's true when the console is a terminal, it might still benefit from a rename to something like `IsTerminal` and documented as "Gets a value indicating whether or not the console is a TTY or PTY interface"

patriksvensson:
* Moves ColorSystem from Profile to Capabilities
* Renames Tty to IsTerminal
* Adds IAnsiConsoleOutput to make output more flexible

Closes #343
Closes #359
Closes #369


patriksvensson:
@phil-scott-78 I think this one is ready for review

phil-scott-78 (approved):

phil-scott-78:
This all looks good to me, and I ran it through its paces with my apps and the examples. Gonna merge it to get it out there for people to test out before the next release.

PR #376, opened 2021-04-18
Removes automatic registration of settings

### LINKED ISSUE #373: Settings registration of CLI commands
Hi,

I've noticed that settings types for CLI commands are registered in the container:

https://github.com/spectreconsole/spectre.console/blob/3e2eea730bf0f6fe4a5e336faa312b526d351079/src/Spectre.Console/Cli/Internal/Extensions/TypeRegistrarExtensions.cs#L36

However, settings are instantiated using `Activator.CreateInstance`:

https://github.com/spectreconsole/spectre.console/blob/3e2eea730bf0f6fe4a5e336faa312b526d351079/src/Spectre.Console/Cli/Internal/Binding/CommandConstructorBinder.cs#L38

Therefore I don't see a reason why the settings types are registered at all.

In my case, I am using Microsoft's generic host, and have enabled [registration validation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceprovideroptions.validateonbuild?view=dotnet-plat-ext-5.0). The validation fails because the container doesn't know how to resolve the registered `ExplainCommand.Settings` because of its c'tor parameters (`string[]? commands, bool? detailed, bool includeHidden`).

My suggestion is to remove the registration of commands' settings. If you agree, I'd be happy to provide a PR.

phil-scott-78:
Original intention was to register Settings automatically for DI. This ran into problem when the container verifies the configuration is valid for a settings using the constructor for initialization. It tries to resolve the parameters and fails.

This removes the automatic registration and falls back ActivatorCreateInstance when no Setting is registered.

Closes #373

### DIFF
--- a/src/Spectre.Console.Tests/Unit/Cli/CommandAppTests.cs
+++ b/src/Spectre.Console.Tests/Unit/Cli/CommandAppTests.cs
@@ -358,15 +358,15 @@ namespace Spectre.Console.Tests.Unit.Cli
 
             // Then
             registrar.Registrations.ContainsKey(typeof(ICommand)).ShouldBeTrue();
-            registrar.Registrations.ContainsKey(typeof(DogSettings));
             registrar.Registrations[typeof(ICommand)].ShouldContain(typeof(DogCommand));
         }
 
         [Fact]
-        public void Should_Register_Default_Command_Settings_When_Configuring_Application()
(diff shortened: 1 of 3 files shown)


patriksvensson (approved):
LGTM!

PR #377, opened 2021-04-18
Add new test framework for consoles

patriksvensson:
This PR adds a more user-friendly test framework for Spectre.Console that can be used by external projects.


patriksvensson:
@phil-scott-78 This one is ready for review whenever you have some spare time.

src/Spectre.Console.Testing/Extensions/StyleExtensions.cs
```diff
@@ -1,7 +1,17 @@
...
     public static class StyleExtensions
     {
+        /// <summary>
+        /// Sets the foreground or background color of the specified style.
+        /// </summary>
+        /// <param name="style">The style.</param>
+        /// <param name="color">The color.</param>
+        /// <param name="foreground">Whether or not to set the foreground color.</param>
+        /// <returns>The same instance so that multiple calls can be chained.</returns>
         public static Style SetColor(this Style style, Color color, bool foreground)
```
phil-scott-78, line 15: Since this is going public, I'd propose making this an enum rather than a bool. With our code it's only being used a parameter in tests, but from a consumer side they might end up with something like this instead

```csharp
style.SetColor(Color.Red, true).SetColor(Color.Blue, false)
```
Other option is maybe just move it to the Tests project and not make it something distributed.

phil-scott-78 (approved):
Everything looks good here. Just a couple of comments on the extension methods off of Style. They are pretty specific to just our tests it feels like so maybe not expose them in something that will be in a library.

src/Spectre.Console.Testing/Extensions/StyleExtensions.cs
thoemmi: I agree that the `boolean` might make reading the code harder. However, I'd think that the odds for the enum having more than two members are pretty low. How about splitting the method into two, `SetForegroundColor` and `SetBackgroundColor`?

patriksvensson: @thoemmi The reason for this is to not have any logic in the test, where the arguments are a boolen provided by a `InlineData` and `Theory`.

However, this should not have been moved to the testing library in the first place. I've removed it so should be ready to merge.

PR #382, opened 2021-04-22
Fix maxWidth bug in Panel

Frassle:
Panel could pass a maxWidth to the child Render function that was greater than the maxWidth that was passed to the Panel render function.
I noticed this when trying to nest a table in a panel and the word wrapping was being disabled.

N.B It might be that Table.Measure ought to not return a measurment larger than maxWidth, but this Panel change seems defensivly sensible regardless.


Numpsy:
> I noticed this when trying to nest a table in a panel and the word wrapping was being disabled.
> 

Hi,
fwiw, I just came here to see about reporting a similar issue (I was trying to set up a panel that contained a table and found that both the nested table and a BreakdownChart that were contained it in had messed up layouts), and if I try a build that has this change in it it seems to be ok :-)

patriksvensson:
@Frassle I see no problem with the fix, but there are build errors that need to be fixed before reviewing and merging.

Frassle:
> @Frassle I see no problem with the fix, but there are build errors that need to be fixed before reviewing and merging.

Yes I'm trying to understand them. 

```
Error: /home/runner/work/spectre.console/spectre.console/src/Spectre.Console.Tests/Unit/TableTests.cs(485,31): error CS0246: The type or namespace name 'FakeConsole' could not be found (are you missing a using directive or an assembly reference?) 
```

FakeConsole is used in the test method just above the one I added, and this builds locally in Windows and WSL for me. So it's a bit odd.

patriksvensson:
@Frassle Have you rebased against the latest upstream main? We removed the `FakeConsole` class and all references a couple of days ago in favor of a new test framework.

Frassle:
Ah and it's testing the merge commit. I had my git remotes in a muddle and thought I was already upto date with this repo but was just upto date with my fork. Rebasing now.

patriksvensson (approved):
LGTM!

patriksvensson:

Issue #385, opened 2021-04-23
Optionally include stack trace when rendering exception


gitfool:
**Is your feature request related to a problem? Please describe.**
[CommandApp.GetRenderableErrorMessage](https://github.com/spectreconsole/spectre.console/blob/91f910326c1e0143fa9dae2602f935d102f29452/src/Spectre.Console/Cli/CommandApp.cs#L122) renders the exception message _without_ the stack trace.

**Describe the solution you'd like**
_Optionally_ include stack trace; ideally in a "pretty" format too.

**Describe alternatives you've considered**
Setting [PropagateExceptions](https://github.com/spectreconsole/spectre.console/blob/91f910326c1e0143fa9dae2602f935d102f29452/src/Spectre.Console/Cli/CommandApp.cs#L108) is a workaround but the result is still a little ugly as the pretty message does not end with a newline.

phil-scott-78:
I'm not sure we'll expose the default error message display with a stack trace, even optionally. Not too many security audits are going to pass with that kind of information leaking so that feels like a non-starter. BUT that leads to the next problem that you pointed out - if someone wants to do that, we are making it tough.

Given a simple try catch I'd expect this to just output the pretty format of the exception

```csharp
try
{
    await app.RunAsync(args);
}
catch (Exception e)
{
    AnsiConsole.WriteException(e);
}
```

My output is 
![image](https://user-images.githubusercontent.com/2447331/115814981-1e2b3e00-a3c4-11eb-9577-84e5c288531d.png)

Which includes that extra "Error: uh oh" before we output the exception. 

Feels like if we are throwing the exception that should happen before the call to print the results of `GetRenderableErrorMessage` and let the user handle the output entirely.

gitfool:
Being optional, and defaulting to false, is safer but still let's the user of the library control their own fate. 😛 

Yeah that's what I meant by it's ugly when the exception propagates. `AnsiConsole.WriteException` looks really nice! Does it play nice if the terminal doesn't support ansi? I'd be happy if an optional flag uses that if set, then I wouldn't propagate exceptions anyway.

patriksvensson:
@gitfool Yes, `AnsiConsole.WriteException` works in non-ansi terminals as well.

phil-scott-78:
@gitfool, the change where we aren't displaying the error when `PropagateExceptions` is now in the latest preview build.

[...]

PR #386, opened 2021-04-23
Removes printing of error messages when propagating

phil-scott-78:
Per #385 if the user wants to propagate the exception and display the full stack trace on an unhandled exception, we have already printed out an error message ourselves resulting in both being displayed. This moves the printing of the message after the throws.

### DIFF
--- a/src/Spectre.Console.Testing/Cli/CommandAppTester.cs
+++ b/src/Spectre.Console.Testing/Cli/CommandAppTester.cs
@@ -53,6 +53,15 @@ namespace Spectre.Console.Testing
             }
             catch (T ex)
             {
+                if (ex is CommandAppException commandAppException && commandAppException.Pretty != null)
+                {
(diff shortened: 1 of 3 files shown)


patriksvensson (approved):

phil-scott-78:
whoops, i disabled my auto-running of the tests last night while working on something and missed the mountain of unit tests that are now failing. gonna look at those now.

phil-scott-78:
Bit more involved than I thought. 

End of the day the unit tests were expecting the pretty message to be displayed, but because we were propagating the exception it wasn't receiving it anymore. It kind of felt like a hack at first to make the `Pretty` property public, but then the more I thought about it the more it felt like if I wanted access to it here while propagating then other users might want it too if they are propagating it.

patriksvensson:
@phil-scott-78 I agree. Merged!

PR #394, opened 2021-04-27
Using Statiq processes for tailwind commands in docs

daveaglick:
This will use the Statiq process feature to run `npm install` before the initial execution and `npm run build:tailwind` after each execution of the engine (which means it'll rebuild Tailwind assets when using the `preview` command after each input file change. I also updated the docs readme to match the current state (I.e. removed the reference to a virtual directory since that's no longer the case).

I think something might be off with the Tailwind configuration though. As far as I can tell the Statiq bits are all running as expected, but the `npm run build:tailwind` command doesn't appear to be pruning the file at `output/assets/styles.css`. Unless I'm misunderstanding and the Tailwind command used during development doesn't actually do pruning, in which case it's probably fine. Here's the output I see from the command when running Statiq in preview mode:

image

(ignore the funny unicode characters, I'm guessing that's due to passing process output through a string and is a problem for another day)

### DIFF
--- a/docs/Docs.csproj
+++ b/docs/Docs.csproj
@@ -31,7 +31,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Statiq.Web" Version="1.0.0-beta.26" />
+    <PackageReference Include="Statiq.Web" Version="1.0.0-beta.27" />
     <PackageReference Include="MinVer" PrivateAssets="All" Version="2.3.1" />
   </ItemGroup>
 
--- a/docs/Program.cs
+++ b/docs/Program.cs
@@ -21,6 +21,11 @@ namespace Docs
                 .AddShortcode("EmojiTable", typeof(EmojiTableShortcode))
                 .AddShortcode("Alert", typeof(AlertShortcode))
                 .AddPipelines()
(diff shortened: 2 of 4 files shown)


phil-scott-78:
So there are two processes right now. `build:tailwind` takes the tailwind.css and tailwind.js and builds out the big 4mb file in the assets folder. That's for dev work. It has every possible tailwind class and utility available.

`build` calls the tailwind cli with the production environment set. this looks at all the generated html files and looks at which css classes are actually being used and prunes them out. 

BUT, with tailwind 2.1the [JIT mode has been introduced](https://tailwindcss.com/docs/just-in-time-mode). this should allow us to point at the cshtml files for pruning and generate a smaller css file for development that's also used in production. This will require modification to the tailwind.config.js.

Question for you though @daveaglick - you mentioned there was a way to keep the process alive for a watch. I'm not on a computer for the next few days, but I suspect this would be just a different option for `AddProcess`? I'd think that's the option we'd want to apply here while it's running.

phil-scott-78:
I'm gonna close this PR. I've taken @daveaglick's branch and added a few things to get it working with tailwind's JIT mode that I want to come along with it.

PR #400, opened 2021-05-02
deleted remaining `1.cs files

cryptocompress:
Deleted remaining "`1.cs" files and added them to .gitignore


patriksvensson:
@cryptocompress Is there a specific reason you want to delete these?

The `'1`  is simply an annotation we use that the file contains a class with one generic type argument, and deleting these files would remove that functionality.

cryptocompress:
Oh, I see. My console (bash) interprets backticks as a command, so i wanted to get rid of them. Looking through the code I thought those are just stale dublicate backups by some editor, as the same methods were just reordered in there. That confused me so much that I overlooked the generic definitions.

Does it make sense to delete the non-generic abstract classes then? Non-generic `AsyncCommand` seems unused.

Or maybe let them extend the generic ones, to not to maintain the code twice?
`public abstract class AsyncCommand : AsyncCommand<EmptyCommandSettings> { }`

`ICommandLimiter` does not exist as non-generic.
And `CommandApp` is completely different code. Not sure about this.

Thanks for the super lib and your time! :+1:

patriksvensson:
> Or maybe let them extend the generic ones, to not to maintain the code twice?

Then the consumer would get to override an Execute method with `EmptyCommandSettings`. We want to hide that from the consumer.

> `ICommandLimiter` does not exist as non-generic.

Yes, it's main purpose is to make the type system happy when limiting an `ICommand<TSettings>` to a known type implementing `TSetting`.

> And `CommandApp` is completely different code. Not sure about this.

`CommandApp<TCommand>` is only encapsulating a `CommandApp` and setting `TCommand` as the default command.

> Thanks for the super lib and your time! 👍

You're welcome! :)

cryptocompress:
> Then the consumer would get to override an Execute method with EmptyCommandSettings. We want to hide that from the consumer.

Deleting the non-generic classes would solve the problem more elegant, as only `TSettings` would be left. The customer also does not have to decide which class must now be overwritten. Less choice; less confusion.

> CommandApp<TCommand> is only encapsulating a CommandApp and setting TCommand as the default command.

This makes only sense when generics were slapped on top later? If it looks like tech-debt, swims like tech-debt, and quacks like tech-debt, it probably is. The backticks are just a side effect. Sorry for raising dust. :)

patriksvensson:
I must say that I do not agree with you at all about this being tech debt.
This was a conscious decision I made when designing the API.

cryptocompress:
Hm, maybe I'm overlooking something. Can you please elaborate a bit more on why you keep the non-generic implementations around? What is the benefit for me, as your customer, to implement a non-generic concrete AsyncCommand for example? When do I want this?
It seems so much easier to implement `AsyncCommand<TSettings>` as you already provide me `EmptyCommandSettings` anyway. What am I missing?

phil-scott-78:
End of the day there are users that are using the non-concrete versions, many times based on code examples we provided in the past. This would be a breaking change for them so killing it off isn't a trivial task for us. Someone might just be upgrading to get a new option on a progress bar and start getting compiler errors in a totally different part of their application unexpectedly. 

It's a conversation perhaps worth having, but not a decision to take lightly. If we do go down this route we certainly wouldn't just delete the code but rather provide a migration path for those users.

cryptocompress:
Totally agree. A breaking change is a very strong argument and requires careful consideration. That's why the proposed solution would be non-breaking:
`public abstract class AsyncCommand : AsyncCommand<EmptyCommandSettings> { ...abstractly override the methods... }`
If it is a breaking-change, please correct me?

This would not solve the backtics issue, as we cannot just delete the files but a simple rename would be non-breaking too.

All of this is just nice to have and by no means important, so feel free to ignore it. But now that I'm here, I'm happy to talk about it and implement it too, if we find a proper solution.

patriksvensson:
Breaking changes doesn't only mean breaking source compatibility. The proposal above would break the ABI and affect upstream users such as Statiq.

I also want to press the fact that this is how I wanted the API to look like when I wrote the library, and I don't see us changing this.

cryptocompress:
Then all that's left is for me to offer you a beer if you ever change your mind. :)

thoemmi:
A small side note: I used to name my files containing generic classes `MyClass´1.cs` too. However, those names confused my colleagues, and they pointed me to a convention, which is also used at Microsoft: they name the files `MyClassOfT.cs`. Maybe this would be an option for Spectre.Console too.

cryptocompress:
done: https://github.com/spectreconsole/spectre.console/pull/406

Issue #405, opened 2021-05-07
Environment Variable Handling


phantlantis:
**Is your feature request related to a problem? Please describe.**
Common rules of precedence dictates that cli params usually override environment variables, but when I attempt to add some options into the Constructor Initialization piece:
```csharp
public class CreateTicketSettings : CommandSettings
{
    public CreateTicketSettings(string platform)
    {
        var platformEV = System.Environment.GetEnvironmentVariable("PLATFORM");
        if (platform != null)
        {
            Platform = platform;
         }
        else Platform = platformEV;
    }
   
   [Description("Platform to create request for (web, ios, android)")]
   [CommandArgument(0, "<platform>")]
   public string Platform { get; set; }
}
```
I can only ever successfully resolve the type (avoiding "Error: Could not resolve type") if I set it to a required argument only with the angle brackets, but this fails validation because I didn't set the parameter in the CLI.

**Describe the solution you'd like**
Ideally, a more "spectre-native" approach to handling environment variables along with parameters - being able to incorporate them and bind them to specific arguments and options and continue to respect the "required" and "optional" settings within spectre. 

**Describe alternatives you've considered**
@patriksvensson  proposed a good solution for using the Validate() method as a workaround [here.](https://github.com/spectreconsole/spectre.console/discussions/402#discussioncomment-708426)

**Additional context**

phil-scott-78:
@phantlantis, PR #407 is merged in. If you take a look at the example from @patriksvensson it uses an Environmental Variable provider. 

Give that a shot if you get a chance.

phantlantis:
@phil-scott-78 , using 0.39.1-preview.0.19 in my project using that sealed class and EnvironmentVariable attribute worked great! Thanks for your help on this!

PR #407, opened 2021-05-08
Add parameter value provider support

patriksvensson:
Adds support for parameter value providers which makes it possible to set custom values for parameters.

For example, #405 could be solved by an environment value provider like this:

```csharp
public sealed class EnvironmentVariableAttribute : ParameterValueProviderAttribute
{
    private readonly string _name;

    public EnvironmentVariableAttribute(string name)
    {
        _name = name ?? throw new ArgumentNullException(nameof(name));
    }

    public override bool TryGetValue(CommandParameterContext context, out object? result)
    {
        // Parameter is a string?
        if (context.Parameter.ParameterType == typeof(string))
        {
            // No value provided?
            if (context.Value == null)
            {
                result = Environment.GetEnvironmentVariable(_name);
                return result != null;
            }
        }

        result = null;
        return false;
    }
}
```

Which would be uses like this:

```csharp
public sealed class MySettings : CommandSettings
{
    [CommandArgument(0, "[MYSETTING]")]
    [EnvironmentVariable("MY_ENV_SETTING")]
    public string MySetting { get; set; }
}
```

cc @phantlantis

### DIFF
--- a/src/Spectre.Console.Tests/Data/Validators/EvenNumberValidatorAttribute.cs
+++ b/src/Spectre.Console.Tests/Data/Validators/EvenNumberValidatorAttribute.cs
@@ -11,19 +11,19 @@ namespace Spectre.Console.Tests.Data
         {
         }
 
-        public override ValidationResult Validate(ICommandParameterInfo info, object value)
+        public override ValidationResult Validate(CommandParameterContext context)
         {
-            if (value is int integer)
+            if (context.Value is int integer)
             {
                 if (integer % 2 == 0)
                 {
                     return ValidationResult.Success();
                 }
 
-                return ValidationResult.Error($"Number is not even ({info?.PropertyName}).");
+                return ValidationResult.Error($"Number is not even ({context.Parameter.PropertyName}).");
             }
 
-            throw new InvalidOperationException($"Parameter is not a number ({info?.PropertyName}).");
+            throw new InvalidOperationException($"Parameter is not a number ({context.Parameter.PropertyName}).");
         }
     }
 }
--- a/src/Spectre.Console.Tests/Data/Validators/PositiveNumberValidatorAttribute.cs
+++ b/src/Spectre.Console.Tests/Data/Validators/PositiveNumberValidatorAttribute.cs
@@ -11,19 +11,19 @@ namespace Spectre.Console.Tests.Data
         {
         }
 
-        public override ValidationResult Validate(ICommandParameterInfo info, object value)
+        public override ValidationResult Validate(CommandParameterContext context)
         {
-            if (value is int integer)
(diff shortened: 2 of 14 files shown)


phil-scott-78:
I was working through the code, and I think I found a bug with both the ValueProvider and TypeProviders together. I'm guessing that the type provider just isn't getting called with the `ValueProvider` path, but I haven't had a chance to fully dig in. Wanted to get my finding written down before I get dragged away from the PC.

Changed up the `ValueProviderSettings` in the tests project to also have a converter that converts the passed in value to a hex string. 

```csharp
public sealed class ValueProviderSettings : CommandSettings
{
    [CommandOption("-f|--foo <VALUE>")]
    [TypeConverter(typeof(HexConverter))]
    [IntegerValueProvider(32)]
    public string Foo { get; set; }
}
```

For the two unit tests the converter runs appropriately when a value is passed via the command line (`settings.Foo.ShouldBe("C")`) in `Should_Not_Override_Value_If_Value_Was_Specified` but I get `null` when `Should_Use_Provided_Value_If_No_Value_Was_Specified` instead of the expected `20`

Here's my converter for what it's worth.

```csharp
public sealed class HexConverter : TypeConverter
{
    public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
    {
        return value is string stringValue && int.TryParse(stringValue, out var intValue)
            ? intValue.ToString("X")
            : base.ConvertFrom(context, culture, value);
    }
}
```

patriksvensson:
@phil-scott-78 The value provider will not return a value, since the target parameter is a string.

```csharp
public override bool TryGetValue(CommandParameterContext context, out object result)
{
    if (context.Parameter.ParameterType == typeof(int))
    {
        if (context.Value == null)
        {
            result = _value;
            return true;
        }
    }

    result = null;
    return false;
}
```

This is fixable, so we use the assigned type converter on the returned value from the value provider, but that means that the HexConverter must be fixed as well.

```csharp
public sealed class HexConverter : TypeConverter
{
    public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
    {
        if (context.PropertyDescriptor.PropertyType != typeof(string))
        {
            throw new InvalidOperationException("Target property must be a string");
        }

        if (value is int integer)
        {
            return integer.ToString("X");
        }

        return value is string stringValue && int.TryParse(stringValue, out var intValue)
            ? intValue.ToString("X")
            : base.ConvertFrom(context, culture, value);
    }
}
```

I will submit a PR for this.

patriksvensson:
@phil-scott-78 I've updated the tests and added type converter support for value providers.

phil-scott-78 (approved):

PR #418, opened 2021-05-17
Add support for hierarchical list prompts

### LINKED ISSUE #412: Multiple selection with groups

It would be awesome to be able to do multiple selections with groups. Something like:

```
[ ] My group
  [ ] Item 1
  [ ] Item 2
[ ] My group 2
  [ ] Another one
  [ ] Yet another one
```

And be able to group and select in groups, etc

patriksvensson:
This PR introduces support for hierarchical list prompts.

There is one breaking change here. Items in multi-selection prompts can no longer be configured to be pre-selected using an index since this doesn't make sense anymore.

prompt_tree

Closes #412


phil-scott-78 (commented):
Code looks good, but I'm going back and forth with the `Hierarchical` property name. Without actually running the code it wasn't terribly clear what would happen between setting it vs `NonHierarchical`. Hierarchical to me indicates that there is going to be a ranking, but nothing that the parents being selectedable.

I kind of like how Vuetify names these two modes "leaf" and "independent" where leaf mode you can only select the leafs vs independent which allows parents to be selected too.

patriksvensson:
@phil-scott-78 That's some good feedback. I've added a `SelectionMode` enum that contains `Leaf` and `Independent` like Vuetify has, which I think makes sense. I've also added selection mode support for `SelectionPrompt` as well.

phil-scott-78 (approved):

examples/Console/Prompt/Program.cs
```diff
@@ -11,6 +13,26 @@ public static void Main(string[] args)
                 return;
             }
 
+            var favorites = AnsiConsole.Prompt(
```
phil-scott-78, line 16: I missed this when doing the first pass on the review. I suspect we didn't want this added.

patriksvensson: You didn't miss it. I accidentally left it in the latest update 😅

patriksvensson: @phil-scott-78 Removed!

PR #423, opened 2021-05-20
Added live display support

### LINKED ISSUE #316: Progress flickers and breaks when there are too many tasks
**Information**
 - OS: Windows
 - Version: 0.38.0
 - Terminal: VS Code, cmd, Terminal (reproducible everywhere)

**Describe the bug**
When using the progress widget, the output starts flickering erratically after some point. My uneducated guess is that the point where it happens is when the tasks stop fitting on the screen. Additionally, scrolling up or down while the progress is ongoing completely breaks the output, causing the output to get duplicated.

Here's a video showcasing what happens (wait till the end to see the whole output without flickering):

https://user-images.githubusercontent.com/1935960/112224493-041b0600-8c34-11eb-9b02-c99561f94fc0.mp4

**To Reproduce**

Here's the code used in above video:

```csharp
using System.Threading;
using System.Linq;
using System.Threading.Tasks;
using System;
using Spectre.Console;

namespace SpectreTest
{
    class Program
    {
        static async Task Main(string[] args)
        {
            var random = new Random();
            using var semaphore = new SemaphoreSlim(5, 5);

            await AnsiConsole.Progress().StartAsync(async ctx =>
            {
                await Task.WhenAll(
                    Enumerable.Range(0, 200).Select(async i =>
                    {
                        await semaphore.WaitAsync();

                        var progress = ctx.AddTask("Task " + i);
                        await Task.Delay(random.Next(100, 3000));
                        progress.Value = progress.MaxValue;

                        semaphore.Release();
                    })
                );
            });
        }
    }
}
```

**Expected behavior**
There should be no flickering and the output should not break if the user scrolls up or down.

**Additional context**
This issue is partially mitigated if I use `HideCompleted(true)` because then the completed tasks don't get displayed and the list is small. However, this also depends on the capacity of semaphore, which is configured in my app by the user.

Even with `HideCompleted(true)` there seem to be issues when the user scrolls up and down. For example, if you write to the console something before you start reporting progress, that text may get overwritten.

If the problem is too difficult to solve, it may also be possible to somehow limit the number of tasks shown on screen, perhaps by having a "...and 57 other" or something similar in the bottom. Ideally, non-completed tasks would show up at the top.

### LINKED ISSUE #415: Multiline WriteLine causes artifacts with Progress, Status
**Information**
 - OS: Windows 10
 - Version: 0.39.0
 - Terminal: Powershell, Windows Terminal

**Describe the bug**
It appears that AnsiConsole.WriteLine (and MarkdownLine, etc.) doesn't handle lines that are longer than the console width when a status or progress is also rendered.

**To Reproduce**
Here's an example the demonstrates the problem:
```c#
using System.Linq;
using Spectre.Console;

AnsiConsole.Status()
           .Start("This is a really long status message that will get written over", ctx =>
           {
               foreach (var width in Enumerable.Range(1, 200))
               {
                   AnsiConsole.WriteLine(string.Join("", Enumerable.Repeat('x', width)));
               }
           });
```

**Expected behavior**
WriteLine should prevent rendered status/progress bar from getting overwritten and mixed into the output.

**Screenshots**
spectre-line-wrap-bug
_A Thread.Sleep was added in the `foreach` for the screenshot above, otherwise the same code_

**Additional context**
- A workaround is to chunk output text by console width and print each.

patriksvensson:
Also adds overflow resolution to LiveRenderable that should fix some problems related to vertical overflow.

livedisplay

live_cropped

Closes #316
Closes #415


src/Spectre.Console/Widgets/Live/LiveDisplayRefreshThread.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        private void Run()
+        {
+            _running.Set();
+
+            try
+            {
+                while (!_stopped.WaitOne(0))
+                {
+                    Thread.Sleep(_refreshRate);
```
patriksvensson, line 48: Remove this

src/Spectre.Console/Widgets/Live/LiveDisplay.cs
```diff
@@ -0,0 +1,146 @@
...
+
+                T result;
+
+                try
+                {
+                    using (new RenderHookScope(_console, renderer))
+                    {
+                        if (AutoRefresh)
+                        {
+                            using (var thread = new LiveDisplayRefreshThread(context, TimeSpan.FromMilliseconds(250)))
```
patriksvensson, line 124: Make refresh rate configurable

src/Spectre.Console/Widgets/Live/LiveDisplay.cs
```diff
@@ -0,0 +1,146 @@
...
+        public VerticalOverflowCropping Cropping { get; set; } = VerticalOverflowCropping.Top;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LiveDisplay"/> class.
+        /// </summary>
+        /// <param name="console">The console.</param>
+        /// <param name="target">The target renderable to update.</param>
+        public LiveDisplay(IAnsiConsole console, IRenderable target)
+        {
+            _console = console;
```
patriksvensson, line 45: Null check

phil-scott-78 (approved):

PR #430, opened 2021-05-24
Add support for fake input for asciicast recording

patriksvensson:
* Also fixes a bug with `SelectionPrompt` and page size
* Allow `IAnsiConsoleInput.ReadKey` to return `null`


phil-scott-78 (approved):
i love it. this is exactly what I envisioned when I started faking the input and it works so much better.

PR #437, opened 2021-05-25
Do not use Thread.Sleep in examples when running on CI

phil-scott-78:
Wanted to see how fast PR builds would be without the `Thread.Sleep` calls in the examples. Looking for a `CI` environmental variable and if so skipping the `Thread.Sleep`

### DIFF
--- a/examples/Console/Live/Program.cs
+++ b/examples/Console/Live/Program.cs
@@ -20,7 +20,7 @@ namespace Spectre.Console.Examples
                     {
                         action();
                         ctx.Refresh();
-                        Thread.Sleep(delay);
+                        ThreadSleeper.Sleep(delay);
                     }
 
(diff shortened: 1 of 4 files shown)


phil-scott-78:
Seems a few minutes shaved. Not sure it's worth it to complicate the demos though

patriksvensson:
@phil-scott-78 So, what should we do with this PR? Should I merge it?

bddckr:
My 2 cents as a user of this fantastic project: The complexity of just conditionalizing the thread sleeps is fine. It also can act as a nice reminder that my CLI projects can be written in a way that allows for a proper test suite that runs in seconds. 😄

patriksvensson:
Closing this for now. Feel free to reopen!

PR #445, opened 2021-06-07
Check for default value before writing property values

### LINKED ISSUE #422: [CLI] Argument array overridden with `null` if no arguments provided
**Information**
 - OS: Win 10
 - Versions: 
   - 0.37.1-preview.0.39, 
   - 0.39.0, 
   - 0.39.1-preview.0.21
 - Terminal: Windows Terminal

**Describe the bug**
When I use argument array I initialize it with `Array.Empty<T>` so I don't have to handle nulls in logic.

```c#
public class Settings: CommandSettings
{
	[CommandArgument(0, "[NAME...]")]
	public string[] Names { get; set; } = Array.Empty<string>(); // I'm initializing it with empty array...
}
```

**To Reproduce**

```c#
public class MyCommand: Command<MyCommand.Settings>
{
	public class Settings: CommandSettings
	{
		[CommandArgument(0, "[NAME...]")]
		public string[] Names { get; set; } =
			Array.Empty<string>(); // I'm initializing it with empty array...
	}

	public override int Execute(CommandContext context, Settings settings) =>
		settings.Names is null 
			? -1 // ...but it gets overriden with null 
			: settings.Names.Length;
}

static class Program
{
	static void Main(string[] args)
	{
		var app = new CommandApp();
		app.Configure(c => c.AddCommand<MyCommand>("do"));
		
		var result3 = app.Run(new[] { "do", "a", "b", "c" });
		Debug.Assert(result3 == 3);
		
		var result0 = app.Run(new[] { "do" });
		Debug.Assert(result0 == 0); // -1 vs 0
	}
}
```

**Expected behavior**
If not arguments are given (`app.Run(new[] { "do" });`) I expect `settings.Names` to be `Array.Empty<string>` not `null`. 

**Actual behavior**
When I give 3 arguments (`app.Run(new[] { "do", "a", "b", "c" });`) I see `3` returned as result. Good. 
When no arguments are given `settings.Names` is overriden with `null`.
It is a regression as it was working as expected in `0.37.1-preview.0.39` but not anymore in `0.39.0` nor `0.39.1-preview.0.21`.

**Screenshots**
N/A

**Additional context**
[spectre-null-fields.zip](https://github.com/spectreconsole/spectre.console/files/6514766/spectre-null-fields.zip)

phil-scott-78:
Previous version was overwriting values that might have been set via a property initializer.

Closes #422

cc @MiloszKrajewski

### DIFF
--- a/src/Spectre.Console.Tests/Data/Settings/OptionalArgumentWithDefaultValueSettings.cs
+++ b/src/Spectre.Console.Tests/Data/Settings/OptionalArgumentWithDefaultValueSettings.cs
@@ -15,6 +15,12 @@ namespace Spectre.Console.Tests.Data
     {
         [CommandArgument(0, "[NAMES]")]
         public string[] Names { get; set; } = Array.Empty<string>();
+
+        [CommandOption("-c")]
+        public int Count { get; set; } = 1;
+
+        [CommandOption("-v")]
+        public int Value { get; set; } = 0;
     }
 
     public sealed class OptionalArgumentWithDefaultValueAndTypeConverterSettings : CommandSettings
--- a/src/Spectre.Console.Tests/Data/Settings/OptionalArgumentWithDefaultValueSettings.cs
+++ b/src/Spectre.Console.Tests/Data/Settings/OptionalArgumentWithDefaultValueSettings.cs
@@ -1,3 +1,4 @@
+using System;
 using System.ComponentModel;
 using Spectre.Console.Cli;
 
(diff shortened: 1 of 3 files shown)


phil-scott-78:
It occurred to me at 3am last night that this was too naive. If the initializer sets it to 1 and the command line setting is 0, ten I don't think we'll override. 

Gonna revisit this tonight if I get a minute unless someone else wants to take a stab at it before then.

patriksvensson:
Would be great if we could add a unit test to catch that regression 😀

phil-scott-78:
First thing I'll write 😃

phil-scott-78:
poured myself a glass of bourbon, got to writing my unit test as a starting point and everything worked. I was worried we'd be sending `default` rather than `null` for the value types, but it looks like my fear was unfounded.

Still nice to have the tests cover this scenario for the future.

Feel free to squash these two commits together if things look good.

patriksvensson (approved):
LGTM! 👍

PR #462, opened 2021-06-21
Added a TypeRegistrar to CommandAppTester

nils-a:
exposed as a public property.
For convenience, and to keep the similarity to the real CommandApp
it is also available in the ctor of CommandAppTester.

This is in regard to #461 

I'm not quite fond of the used-in-this-one-place implementation of `FakeRegistrar`, but I see no real good alternative except exposing the `DefaultRegistrar` to be used in tests or creating a second (fully implemented) `ITypeRegistrar` for tests.

### DIFF
--- a/src/Spectre.Console.Testing/Cli/CommandAppTester.cs
+++ b/src/Spectre.Console.Testing/Cli/CommandAppTester.cs
@@ -12,6 +12,20 @@ namespace Spectre.Console.Testing
         private Action<IConfigurator>? _configuration;
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="CommandAppTester"/> class.
+        /// </summary>
+        /// <param name="registrar">The registrar.</param>
+        public CommandAppTester(ITypeRegistrar? registrar = null)
+        {
+            Registrar = registrar;
+        }
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


patriksvensson:
@nils-a I just pushed a fix for the analyzer where we moved the test project to the root. Sorry, didn't think about your PR when I did that so you will have to rebase your branch and fix the merge conflict 😢 

I think the fake registrar is OK but perhaps we could re-use the [FakeTypeRegistrar](https://github.com/spectreconsole/spectre.console/blob/main/test/Spectre.Console.Tests/Utilities/FakeTypeRegistrar.cs) for this purpose?

nils-a:
@patriksvensson 
> so you will have to rebase your branch and fix the merge conflict 

Not a problem. 👍 

> perhaps we could re-use the FakeTypeRegistrar for this purpose?

I've thought about that. (In fact I implemented it, then reverted..) The problem I see with that approach is that `FakeTypeRegistrar` currently returns `null` on `Build()`. Implementing an `ITypeResolver` would be a breaking change in *Spectre.Console.Testing*.

But... Thinking about it now... I could add the ITypeResolver as optional feature. Have the result of the `Build` method injectable per `Func`.

patriksvensson (approved):
LGTM!

PR #463, opened 2021-06-22
The first analyzers

phil-scott-78:
Handful of analyzers, and a new project that *hopefully* should be packaged up properly as an analyzer.


patriksvensson:
@phil-scott-78 Yay! 🎉 
Will take a look at this tonight!

src/Spectre.Console.Analyzer/Analyzers/BaseAnalyzer.cs
```diff
@@ -0,0 +1,25 @@
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Spectre.Console.Analyzer
+{
+    /// <summary>
+    /// Base class for Spectre analyzers.
+    /// </summary>
+    public abstract class BaseAnalyzer : DiagnosticAnalyzer
```
patriksvensson, line 8: Could we perhaps rename this to something like `SpectreAnalyzer` or similar? Not a big issue per se, but makes it easy to understand that it's part of the solution.

examples/Console/AnalyzerTester/AnalyzerTester.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
patriksvensson, line 1: Since it's not really an example, should we move this project to `src` and call it something like `Spectre.Console.Analyzer.Sandbox`, `Spectre.Console.Analyzer.TestBed` or similar?

patriksvensson (approved):
LGTM! Got some minor things, but I'll let you decide about it.

phil-scott-78:
both of those changes sounded like a good idea to me. Made the changes and pushed them up.

patriksvensson:
Merged! 🎉

PR #469, opened 2021-06-25
Adds documentation for analyzers

phil-scott-78:
Adds documentation along with a template, layout, and index page for the analyzers. Also updates the analyzer to point to the url these rules will live at.


patriksvensson (approved):
LGTM!

PR #484, opened 2021-07-13
Add blog to docs

phil-scott-78:
I want to checkout the output generated from the CI and make sure things don't get squirelly with github actions and playwright. But if that works this might just be ready. 

I'm still not 100% pleased with the index page for the categories, but they still look ok there and aren't going to be very frequently traveled pages anyways.


phil-scott-78:
Cleaned up some CI errors that popped up then verified the artifact that the docs look good. 

Might be some spelling mistake and the first post isn't exactly pulitzer award winning material, but I feel good about the tech side of things at least :-)

.github/workflows/ci.yaml
```diff
@@ -21,7 +21,8 @@ jobs:
     - name: Setup dotnet
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 5.0.301
+        dotnet-version: 6.0.x
```
patriksvensson, line 24: I think this change will need to be done in `publish.yaml` as well

patriksvensson (changes requested):
Looks good to me, but found one thing that might need to be fixed

.github/workflows/ci.yaml
phil-scott-78: oh great catch. got that updated to 6.0.x with the latest push

patriksvensson (approved):
LGTM 👍

patriksvensson:
Merged! 🎉

PR #516, opened 2021-08-20
Automatically provide global using

### LINKED ISSUE #515: Automatically provide global using for Spectre.Console
It's (almost?) sure that someone referencing the package will right-away use its provided types.

With the new C# 10 global usings feature, it would be great if referencing the package automatically provided such global usings, so that a simple console top-level program could right-away be just:

```
AnsiConsole.WriteLine("Hello World!");
```

No added usings needed.

kzu:
The new C# 10 feature lights up automatically via the provided item group with the `Import` element.

There's talk of a future opt-out (or opt-in?) MSBuild property to control these global usings, but the provided approach will work and integrate seamlessly with whatever the project/SDK configures.

For reference, see https://github.com/dotnet/aspnetcore/issues/32451.

Note we provide the "legacy" MSBuild xmlns attribute so as not to break downlevel IDEs or legacy project files that happen to reference this package.

Fixes #515

### DIFF
--- a/src/Spectre.Console/Spectre.Console.props
+++ b/src/Spectre.Console/Spectre.Console.props
@@ -1,0 +1,7 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <ItemGroup>
+    <Import Include="Spectre.Console" />
+  </ItemGroup>
+
+</Project>
--- a/src/Spectre.Console/Spectre.Console.csproj
+++ b/src/Spectre.Console/Spectre.Console.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>net5.0;netstandard2.0</TargetFrameworks>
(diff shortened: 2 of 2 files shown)


src/Spectre.Console/Spectre.Console.props
```diff
@@ -0,0 +1,7 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <ItemGroup>
+    <Import Include="Spectre.Console" />
```
alexrp, line 4: Shouldn't this be `Using`? See: https://github.com/dotnet/sdk/pull/19599

Also, how about respecting `ImplicitUsings`?

kzu: Yep, should be using for next preview/stable. 

WRT respecting the condition: https://twitter.com/kzu/status/1430001920022831110?s=19

It's rather pointless, but I can add that, sure. 

Before any of that though, would be nice to know if there's even interest in merging this eventually ;-)

alexrp: The point of respecting the condition would be that if the user has set `ImplicitUsings=false`, they probably don't want `Spectre.Console` added to the `Using` item type. It's a way of saying "I don't want implicit usings coming from parts of the build that I can't see".

kzu: That's not how that property works. It turns off global usings from MSBuild items entirely. See  https://twitter.com/kzu/status/1430001920022831110?s=19

alexrp: I'm not seeing what you're seeing? All I see is that it prevents the SDK from adding its default usings. The SDK will still generate a usings file if the `Using` item type is non-empty.

phil-scott-78:
We've let this one hang out seeing where the .net world landed on global usings. Seems devs are handling the imports on their own and  packages aren't including them. Plus tooling is becoming more common to assist those not familiar with the syntax. 

I'm gonna close this PR. Appreciate the thought, but just not a feature we want to push.

PR #520, opened 2021-08-25
Optimize operating system detection.

teo-tsirpanis:
This PR changes the operating system from `RuntimeInformation.IsOSPlatform` to use `OperationSystem.IsXxx`. Their advantage is that they are better optimized by the JIT, and recognized by the trimmer.

Because this class is not available on .NET Standard, it was minimally shimmed in an internal class that gets the correct version of Windows [the way .NET 5+ does it](https://github.com/dotnet/runtime/blob/e0ebe9ee047b07352d262e5f1335e186dc0f1b50/src/libraries/System.Private.CoreLib/src/System/Environment.Windows.cs#L107-L119). As a disadvantage, unsafe code was enabled to write the shim, but is allowed only when targeting .NET Standard 2.0.

### DIFF
--- a/src/Spectre.Console/AnsiConsoleFactory.cs
+++ b/src/Spectre.Console/AnsiConsoleFactory.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Runtime.InteropServices;
 using Spectre.Console.Enrichment;
 using Spectre.Console.Internal;
 
@@ -89,7 +88,7 @@ namespace Spectre.Console
                 if (buffer.IsStandardOut() || buffer.IsStandardError())
                 {
                     // Are we running on Windows?
-                    if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                    if (OperatingSystem.IsWindows())
                     {
                         // Not the first console we're creating?
                         if (AnsiConsole.Created)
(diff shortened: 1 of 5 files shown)


patriksvensson:
@teo-tsirpanis While I appreciate the work you put into this, I do not want to introduce the extra complexity this PR means to the project. Especially since there currently isn't a known performance problem regarding the detection of operating systems. After 1.0 has been released, we will move away from netstandard2.0, and then this change would be considered low risk.

I'm sorry. Like I said, I really appreciate you taking the time to work on this.

For next time, to avoid doing work that we might not be in a position to merge, see our contribution guidelines: https://github.com/spectreconsole/spectre.console/blob/main/CONTRIBUTING.md#get-buyoff-or-find-open-community-issues-or-features

teo-tsirpanis:
Hello and thanks for your feedback. The PR's complexityconcentrated to a single file that can be simply removed once the library stops targeting .NET Standard 2.0, and the rest of the code changes actually _decrease_ complexity; we replace single-parameter method calls with parameterless ones, and use the best possible APIs to detect the running platform for each framework.

Perhaps it is the introduction of unsafe mode and the additional Windows API call that concerns you. Would reverting to the previous OS version detection code that used a regex on `RuntimeInformation.OSDescription` make things better?

Unless of course support for .NET Standard 2.0 is going to be dropped soon either way.

patriksvensson:
@teo-tsirpanis My main concern is that we got something that works right now, and we've had some problems before with OS detection which is stable now, that's my main concern. I would gladly merge this fix once we've dropped netstandard support!

PR #527, opened 2021-08-31
Fix the style parameter nullable annotation on AnsiConsoleExtensions

0xced:
The style parameter is actually nullable. Also, the documentation of the style parameter has been made explicit that `Style.Plain` is used when a `null` style is passed.


patriksvensson (approved):
LGTM! 👍

patriksvensson:

PR #533, opened 2021-09-09
Removed the additional registration of ICommand

nils-a:
As discussed in https://github.com/spectreconsole/spectre.console/discussions/532 the additional registration is not needed anymore.

I modified two tests which were specifically testing for the `ICommand` registration.
I also tested all the demos, and they all worked.

**However**, this will be a breaking change for anyone who implemented a custom `ITypeRegistrar` and relied upon the `ICommands` to be registered.


patriksvensson (approved):

patriksvensson:
Related to https://github.com/spectreconsole/spectre.console/issues/541

Issue #539, opened 2021-09-14
Add EnvironmentVariableAttribute example to provided API


koliyo:
**Is your feature request related to a problem? Please describe.**
I want to use environment variables to populate options.

**Describe the solution you'd like**
This seems to have already been solved in #407 

But instead of copy pasting the `EnvironmentVariableAttribute` it would be nice if it was part of the `Spectre.Console` API

koliyo:
I have a WIP pull request here, but not yet any tests.

https://github.com/spectreconsole/spectre.console/pull/540

Not sure when I will have time to complete the PR, so please use it as a suggested implementation and make a complete implementation if this looks like a good idea...

patriksvensson:
@koliyo It looks like a good idea, but closing the PR until you have time to revisit it.

FrankRay78:
Not enough desire amonst contributors to implement this feature.

@anyone - feel free (at any time) to reopen, explain the need, and submit a complete PR for it.

PR #540, opened 2021-09-14
Draft: Implement EnvironmentVariableAttribute with TypeConverter

koliyo:
This use the  `EnvironmentVariableAttribute` from https://github.com/spectreconsole/spectre.console/discussions/402 and extends the implementation with `TypeConverter` to convert from the environment variable string.

NOTE: This does not yet include testing


patriksvensson:
Closgin for now, until PR is complete. See https://github.com/spectreconsole/spectre.console/issues/539

PR #552, opened 2021-09-27
Adds a segment builder for merging multiple segments

### LINKED ISSUE #551: Low performance rendering big Tree objects
**Information**
 - OS: macOS
 - Version: Big Sur 11.5.2
 - Terminal: iTerm

**Describe the bug**
Rendering big trees is noticeable slow in spectre.console - moreover compared with other tools in which the same rendering seems to happen instantaneously.

**To Reproduce**
I'm playing with the library by creating a clone of the `tree` utility. For this, first i traverse the FileSystem, then build the `Tree` object, then render it using `AnsiConsole.Render`. This is the detailed log of how much time each operation took:

```text
23:23:49.537 FsTree - Started walking tree '/Users/sluisp/wkspaces/codice'
23:23:50.281 FsTree - Finished walking tree. Found 10703 directories and 95 files
23:23:50.287 FsTreeCommand - Started rendering the tree with AnsiConsole
23:23:50.290 RenderTree - Started building Spectre.Console.Tree object
23:23:50.369 RenderTree - Finished building Spetre.Console.Tree object
23:24:35.030 FsTreeCommand - Finished rendering tree
```

- Traversing the FS: 744ms
- Building the `Spectre.Console.Tree` object: 79ms
- Calling `AnsiConsole.Render` with that tree: 84661ms

As you can see in the logs, the tree is comprised of 10798 nodes. It's quite big, but for a FileSystem I would consider it quite standard. Each node is only plain text - no text colors, no tables, not anything fancy. A exact `tree` clone regarding what goes to the screen.

**Expected behavior**
The rendering could be faster. I'm sure lot's of things are happening behind the scenes, but... maybe there is room for improvement :)

Or maybe I'm using the library wrong and I need to adjust a buffer or something like that for it to work faster, but I didn't found anything documented.

phil-scott-78:
When merging a large number of segments together we were doing a tremendous amount of allocation especially related to strings due to concatenation.

This adds an internal `SegmentBuilder` that uses a `StringBuilder` for building up the text rather than creating a new instance and doing a concat operation for each segment.

Here's the perf for a 10,000-node tree.

Before
-----
image

After
----
image

In the before you'll notice we are doing 44gb of memory allocations, mostly strings. Afterwards thanks to the `StringBuilder` this drops to 44mb. Performance obviously comes with this too, as the time to build goes from 32s down to 133ms. Not too shabby!

Unit tests all pass, obviously, but @patriksvensson I definitely could use a careful eye and feedback here because I get nervous anytime I write code that messes with the control characters :-)

Fixes #551

### DIFF
--- a/src/Spectre.Console/Rendering/Segment.cs
+++ b/src/Spectre.Console/Rendering/Segment.cs
@@ -449,36 +449,36 @@ namespace Spectre.Console.Rendering
 
             var result = new List<Segment>();
 
-            var previous = (Segment?)null;
+            var segmentBuilder = (SegmentBuilder?)null;
             foreach (var segment in segments)
             {
-                if (previous == null)
+                if (segmentBuilder == null)
                 {
-                    previous = segment;
+                    segmentBuilder = new SegmentBuilder(segment);
                     continue;
                 }
 
                 // Both control codes?
-                if (segment.IsControlCode && previous.IsControlCode)
+                if (segment.IsControlCode && segmentBuilder.IsControlCode())
                 {
-                    previous = Control(previous.Text + segment.Text);
+                    segmentBuilder.Append(segment.Text);
                     continue;
                 }
 
                 // Same style?
-                if (previous.Style.Equals(segment.Style) && !previous.IsLineBreak && !previous.IsControlCode)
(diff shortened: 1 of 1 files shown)


patriksvensson (approved):
LGTM!

PR #563, opened 2021-09-29
Future-proof conditional compilation

0xced:
* Invert `#if NET5_0` conditions so that when adding net6.0 target framework, the _new_ APIs are used.
* Use `NET5_0_OR_GREATER` instead of `NET5_0` to ensure consistent behaviour on future target frameworks.


patriksvensson (approved):
LGTM! 👍

PR #564, opened 2021-09-29
Introduce MarkupInterpolated and MarkupLineInterpolated extensions

0xced:
These new methods enable easily writing markup with a nice and intuitive syntax without having to worry about escaping the markup.

```csharp
string input = args[0];
string output = Process(input);
AnsiConsole.MarkupLineInterpolated($"[blue]{input}[/] -> [green]{output}[/]");

```

The `Interpolated` suffix was inspired by the Entity Framework Core [FromSqlInterpolated][1] method.

[1]: https://docs.microsoft.com/en-us/ef/core/querying/raw-sql#passing-parameters

### DIFF
--- a/src/Spectre.Console/AnsiConsole.Markup.cs
+++ b/src/Spectre.Console/AnsiConsole.Markup.cs
@@ -28,6 +28,24 @@ namespace Spectre.Console
 
         /// <summary>
         /// Writes the specified markup to the console.
+        /// <para/>
+        /// All interpolation holes which contain a string are automatically escaped so you must not call <see cref="StringExtensions.EscapeMarkup"/>.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// string input = args[0];
+        /// string output = Process(input);
+        /// AnsiConsole.MarkupInterpolated($"[blue]{input}[/] -> [green]{output}[/]");
+        /// </code>
+        /// </example>
+        /// <param name="value">The interpolated string value to write.</param>
+        public static void MarkupInterpolated(FormattableString value)
+        {
+            Console.MarkupInterpolated(value);
+        }
+
+        /// <summary>
+        /// Writes the specified markup to the console.
         /// </summary>
         /// <param name="provider">An object that supplies culture-specific formatting information.</param>
         /// <param name="format">A composite format string.</param>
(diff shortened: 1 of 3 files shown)


patriksvensson:
Is there a related issue for this PR?

phil-scott-78:
@0xced, I've been working on something similar that I wanted to get done with a RFC. It has the same idea you have but takes it a bit further by also adding the ability to format the parameters with styling.

Since you and I seem to be looking in the same direction, think you could be the first to let me know what you think about how this works? 

My POC is here - https://github.com/phil-scott-78/Spectre.Console.StringInterpolation

0xced:
> And if so it kinda feels like including `Interpolated` in the name, while accurate, might be a bit too verbose.

I totally agree with you. I went with the `Interpolated` suffix because this is how Entity Framework Core does it with [FromSqlInterpolated][1] and I don't know of any other example of .NET code taking advantage of the `FormattableString` type, but I haven't really looked.

> Thoughts on changing it to `MarkupFormat` and `MarkupFormatLine`?

I like them better but I'm afraid they sound too much like methods using [composite formatting][3] which is exactly what they are not!

> If we went this route I might also provide overrides to match `string.Format` signatures too for people like F# users that might not want to use the interpolated strings but still want the autoformatting.

I'm not familiar with F# at all and I don't see how your proposal would look like. Care to elaborate?

Anyway, I'm really open to something else, I don't particularly like `Interpolated` either.

<hr>

By the way, can someone with write access to the repository (probably Patrik or yourself) turn this pull request [into a draft][2]?

[1]: https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.relationalqueryableextensions.fromsqlinterpolated
[2]: https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/
[3]: https://docs.microsoft.com/en-us/dotnet/standard/base-types/composite-formatting

nils-a:
> can someone with write access to the repository (probably Patrik or yourself) turn this pull request into a draft?

Done.

phil-scott-78 (approved):
We are going through the backlog and this is one item that we (well, I) let sit too long.

@0xced if you get a chance can you review the conflicts? Or alternatively I can build off of what you have and get it merged

0xced:
I'm away from computer for this week, so please go ahead as you see fit, Phil.

phil-scott-78 (approved):

phil-scott-78:
I'm a bit concerned I broke github by cheating and using their UI to resolve the merge conflict....

PR #569, opened 2021-10-03
Improve the error message when acquiring the interactive semaphore fails

0xced:
Using the same error message as in the sync version of the `Run` method which is indeed much better than "Could not aquire the interactive semaphore".


patriksvensson (approved):

patriksvensson:
Related to https://github.com/spectreconsole/spectre.console/issues/578

patriksvensson:

PR #585, opened 2021-10-06
Obsolete the AnsiConsoleFactory class

0xced:
`AnsiConsoleFactory` should be an internal static class. Creating an `IAnsiConsole` can be achieved through `Spectre.Console.AnsiConsole.Create(AnsiConsoleSettings)`

Since `AnsiConsoleFactory` is public, it can't be changed from a non static class to a static class so obsoleting it is the second best thing to do.

Eventually, when `AnsiConsoleFactory` becomes internal and static, `AnsiConsole.Create` can be simplified to this (and the private static field `_factory` can be removed)

```
public static IAnsiConsole Create(AnsiConsoleSettings settings)
{
    return AnsiConsoleFactory.Create(settings);
}
```


0xced:
I will try to think about opening issues first for discussing.

For this particular case, it was not much trouble, though. If this change is not deemed appropriate, then just close the pull request without merging, I won't be offended. 😉

phil-scott-78:
This is another PR I think we are good with moving forward with no. But because it came before the big global namespace PR it does need to have conflicts resolved

0xced:
Just [as I predicted](https://twitter.com/0xced/status/1457072951346425859). 😜

Anyway, I can have a look next week but if you want to take over this PR earlier, please go ahead, it's trivial if I remember correctly.

nils-a (approved):
LGTM

nils-a:

PR #590, opened 2021-10-14
(#502) Added GetParent and GetParents to MultiSelectionPrompt

### LINKED ISSUE #502: Return group in SelectionPrompt and MultiSelectionPrompt
I would love when using `SelectionPrompt` or `MultiSelectionPrompt` with choice groups, that I could get both the selected item, but also the group it belongs to as output.

Consider this code:

```csharp
var api = AnsiConsole.Prompt(
    new SelectionPrompt<string>()
        .Title("What API do you want to explore?")
        .PageSize(10)
        .MoreChoicesText("[grey](Move up and down to reveal more options)[/]")
        .AddChoiceGroup("Cloud API", new string []
        {
            "Booking API"
        })
        .AddChoiceGroup("Site API", new string[]
        {
            "Booking API"
        }));
```

If I have the same choice in different groups, I would like to know which group it belongs to.

Right now I would have to wrap the strings in my own little class such as:

```csharp
public class Choice
{
    public string Name { get; set; }
    public string Group { get; set; }

    public override string ToString() => Name;
}
```

Would be great if somehow `SelectionPrompt` and `MultiSelectionPrompt` could somehow also return the choice group when returning a value.

nils-a:
So it is possible to find the parent(s) of a given item.

fixes #502

### DIFF
--- a/src/Spectre.Console/Prompts/MultiSelectionPrompt.cs
+++ b/src/Spectre.Console/Prompts/MultiSelectionPrompt.cs
@@ -122,7 +122,7 @@ namespace Spectre.Console
             var promptItem = Tree.Find(item);
             if (promptItem == null)
             {
-                throw new ArgumentOutOfRangeException(nameof(item), "item not found in tree.");
+                throw new ArgumentOutOfRangeException(nameof(item), "Item not found in tree.");
             }
 
             var parents = new List<ListPromptItem<T>>();
--- a/src/Spectre.Console/Prompts/MultiSelectionPrompt.cs
+++ b/src/Spectre.Console/Prompts/MultiSelectionPrompt.cs
@@ -112,6 +112,41 @@ namespace Spectre.Console
                 .ToList();
         }
 
+        /// <summary>
+        /// Returns all parent items of the given <paramref name="item"/>.
+        /// </summary>
+        /// <param name="item">The item for which to find the parents.</param>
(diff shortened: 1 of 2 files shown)


src/Spectre.Console/Prompts/MultiSelectionPrompt.cs
```diff
@@ -112,6 +112,41 @@ public async Task<List<T>> ShowAsync(IAnsiConsole console, CancellationToken can
...
+        /// Returns all parent items of the given <paramref name="item"/>.
+        /// </summary>
+        /// <param name="item">The item for which to find the parents.</param>
+        /// <returns>The parent items, or an empty list, if the given item has no parents.</returns>
+        public IEnumerable<T> GetParents(T item)
+        {
+            var promptItem = Tree.Find(item);
+            if (promptItem == null)
+            {
+                throw new ArgumentOutOfRangeException(nameof(item), "item not found in tree.");
```
patriksvensson, line 125: Nitpick: Capital `I` in `item`

patriksvensson (changes requested):

src/Spectre.Console/Prompts/MultiSelectionPrompt.cs
nils-a: 👍

patriksvensson (approved):

PR #591, opened 2021-10-15
Upgrades Statiq and re-enables docs publishing

phil-scott-78:
Previous versions of Statiq had a bug with .NET 6 RC1. The latest has a workaround that allows publishing


phil-scott-78:
hold off on the merge, I missed something I think with the publish step

PR #610, opened 2021-11-01
Escape any Markup when displaying selected prompt items

nils-a:
If the item contained escaped markup, after the call to RemoveMarkup
the string will contain unescaped markup (that the user explicitly had
escaped before) for those cases we need to escape all remaining markup.


nils-a:
@patriksvensson  this is the fix I talked about and the rather crude test for it.
 
I'm not sure whether https://github.com/spectreconsole/spectre.console/pull/610/files#diff-0d82c633573706ac9b31907271477cf0d7246453131d21370e7430bc877ae7f1R70 is really a good idea.

patriksvensson (approved):
LGTM!

PR #618, opened 2021-11-10
Fix type conversion in the default pair deconstructor implementation

0xced:
The code using the TypeConverter was written backwards (using `ConvertTo` instead of `ConvertFrom`) and would thus throw an exception. For example, when converting an enum:
> System.NotSupportedException: 'EnumConverter' is unable to convert 'System.String' to 'System.DayOfWeek'.

A test covering conversion of enums was added: `Should_Map_Pairs_With_Enum_Value_To_Pair_Deconstructable_Collection_Using_Default_Deconstructor`


nils-a:

Issue #555, opened 2021-09-27
Internal exceptions being thrown


daveaglick:
I updated Spectre.Console and started seeing handled exceptions being thrown, got curious, and traced it back to PR #376. The problem is that while the new `try/catch` introduced in this PR around the `resolver.Resolve(settingsType)` call works, it ends up creating an expected control flow where throwing is "normal". That's fine as far as it goes, but it does result in unnecessary exceptions under the hood which will both impact performance (though likely not in any critical way) and trigger exception filters when debugging. And it feels icky :)

The reason I didn't see these exceptions before the update was because `TypeRegistrarExtensions.RegisterDependencies()` was registering the settings types in my container for me, so when the old code asked the container for them, they were already in there. That code was removed in this PR so the container doesn't have the settings type registered any more.

It's unclear whether this is a "problem" that needs solving. I generally don't love throwing exceptions during expected flows, even if I'm catching them. If there's an interest in changing the behavior the best thing I can think of is to introduce an `ITypeResolver.TryResolve()` that can return a null value for unregistered types without throwing (or alternatively use a `bool`/`out` pattern). Then use that to resolve the settings type without a `try`/`catch` and fallback to direct instantiation if it returns null.

Would you be interested in a PR for this or should the behavior be considered by design?

patriksvensson:
@daveaglick I would want us to do the right thing here, so a PR would be appreciated!

daveaglick:
Will do! Shouldn't be a tough one (famous last words). I may not have a chance to open a PR until Friday for...errr...reasons...

![image](https://user-images.githubusercontent.com/1020407/135321045-6c93751a-690f-432b-a3dc-33d1f3b7260e.png)

👀

nils-a:
@daveaglick you're still working on this one? (I mean, there's still a day left in your self-chosen timeframe, so I wouldn't want to put any pressure on you. 😬)

daveaglick:
Lol. I like to live on the edge. Nothing like a little pressure!

daveaglick:
I was all set to implement this as described (only two months later, nbd!) but after digging in, realized that maybe this is just a documentation issue. Consider this code I found in `CommandConstructorBinder`:

```csharp
var value = resolver.Resolve(parameter.ParameterType);
if (value == null)
{
    throw CommandRuntimeException.CouldNotResolveType(parameter.ParameterType);
}
```

[...]

PR #620, opened 2021-11-13
(#555) Clarify ITypeResolver returns `null`

nils-a:
and does not throw on unresolvable types.
Also changed the `TypeResolverAdapter` to adhere
to those expectations and removed the now no longer
needed try-catch from `CommandPropertyBinder`.


patriksvensson (approved):
LGTM! 👍

PR #625, opened 2021-11-14
Update examples to net6.0

### LINKED ISSUE #604: Update examples to net6.0 TFM
When #602 is implemented and after net6.0 is released in a stable version, we should update the examples to use net6.0 TFM

patriksvensson:
Closes #604


phil-scott-78 (approved):
Looking good!

PR #513, opened 2021-08-20
Fix parsing of exceptions on .NET Framework

0xced:
On .NET Framework, `exception.ToString()` uses a slightly different format than on .NET Core.

So in order to properly transform an `Exception` into an `ExceptionInfo` on both .NET Core and .NET Framework we use `exception.StackTrace` + `exception.InnerException`. As an added benefit, it greatly simplifies the implementation of the `ExceptionParser` class.

This required a few tweaks to the test project to make it compile and run on .NET Framework 4.8 (first commit of this pull request).

The last commit of this pull request addresses a new intermittent test failure that appeared when running tests on .NET Framework.

PR #637, opened 2021-11-24
Simplify stack frame parsing

0xced:
This is the natural extension of #513 which already removed a lot of regex parsing. 

The `ExceptionParser` class is renamed into `ExceptionConverter` because there is no more string parsing performed at all. `Exception` is converted into `ExceptionInfo` and `StackFrame` is converted into `StackFrameInfo`.


patriksvensson (approved):

patriksvensson:

PR #639, opened 2021-11-24
Fix exception formatting for generic methods

0xced:
The generic parameters were double escaped and would display as `[[T0,T1,TRet]]` instead of `[T0,T1,TRet]`. This is because the `builder.AppendWithStyle` method already escapes its value so the caller must not escape the passed value.


nils-a:
@0xced looks reasonable. I have not done any testing of this currently, but I think I will get to it tomorrow. Would you mind opening an issue, so we can track this "correctly"?

0xced:
> Would you mind opening an issue, so we can track this "correctly"?

Done in #641

nils-a (approved):
LGTM!

nils-a:

PR #642, opened 2021-11-28
Added the ability to hide CommandOptions.

### LINKED ISSUE #631: Add the ability to hide CommandOptions
Coming from https://github.com/spectreconsole/spectre.console/discussions/629, it would be good to have the ability to hide `CommandOptions`.

Something like

```cs
public class MySettings : CommandSettings
{
    [CommandOption("--hidden-opt", IsHidden = true)]
    public bool HiddenOpt { get; init; }
}
```

**Additional context**
See #629

twaalewijn:
CommandOptions now has an IsHidden property that, when set to true, will cause the option to be hidden from the following cases:

- Help text using `-h|--help`
- Xml representations generated with the `cli xml` command
- Diagnostics displayed with the `cli explain` command

Hidden options can still be outputted with `cli explain` using the `--hidden` option that is also used to display hidden commands.

Fixes #631

### DIFF
--- a/docs/input/cli/settings.md
+++ b/docs/input/cli/settings.md
@@ -53,6 +53,15 @@ There is a special mode for `CommandOptions` on boolean types. Typically all `Co
 public bool Debug { get; set; }
 ```
 
+### Hidden options
+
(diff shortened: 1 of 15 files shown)


nils-a (approved):
LGTM!

PR #663, opened 2021-12-16
Add release notes for 0.43.0

patriksvensson:


phil-scott-78:
I've read through it and it looks good. 

I'm gonna take a look at why playwright suddenly isn't working later today though

phil-scott-78:
Opps, looks like commit 8ce40bac3c8617b57385f8df3cc16e6b6dd028e0 might be the cause. I'm thinking that the change went from an explicit install to a restore, but without a tool-manifest it doesn't know to install playwright hence the error.

patriksvensson:
@phil-scott-78 Ah, I see. I had to remove it since it errored when installing since it was specified in the manifest. Any suggestions on how to move forward here?

phil-scott-78:
I', thinking `dotnet tool` saw the `dotnet-tools.json` in the root of the repository and put the entry for `microsoft.playwright.cli` there. But we are running these commands from a child folder. So, when we are restoring from the docs folder perhaps it isn't finding it? I'm on my phone so I can test, but perhaps a `dotnet-tools.json` file in the `docs` folder is needed with playwright installed there?

phil-scott-78:
I think I got the build fixed with #667. After that's reviewed and (possibly) merged think we could rebase? Content looks good, just gotta make sure it gets deployed :-)

phil-scott-78 (approved):

PR #665, opened 2021-12-20
docs: Add missing inheritance and nullable types in settings.md

seclerp:
Seems like `settings.md` documentation file is missing some nullable types for non-required properties. Also, examples at the bottom are not compilable because there is a missing inheritance


patriksvensson (approved):

phil-scott-78:
Content looks good. I'm going to go ahead and rebase and merge - build failure was related to a hiccup in the doc build process not the content.

PR #668, opened 2021-12-22
Add global usings

patriksvensson:


phil-scott-78 (approved):
This works for me. Only question is if the file and it's location makes sense for @nils-a so I'll leave it open until he can chime in

nils-a (approved):
Yes. This is brilliant.

PR #681, opened 2022-01-06
Add option to configure the style of the choices list and the default value in TextPrompt

ap0llo:
TextPrompt currently uses a fixed style for the default value (green text) and the list of choices (blue text).

This PR adds two new properties to TextPrompt (`DefaultValueStyle` and `ChoicesStyle`) that allow customozing the style of these values.
If no styles are set, the current green & blue values continue to be used.

Note: I'm unsure if there are tests that cover the styling of TextPrompt. If there is a place I can add a test for this, it would be great if you could point me in the right direction.

### DIFF
--- a/src/Spectre.Console/Prompts/TextPrompt.cs
+++ b/src/Spectre.Console/Prompts/TextPrompt.cs
@@ -69,6 +69,11 @@ public sealed class TextPrompt<T> : IPrompt<T>
     public Style? DefaultValueStyle { get; set; }
 
     /// <summary>
+    /// Gets or sets the style in which the list of choices is displayed.
+    /// </summary>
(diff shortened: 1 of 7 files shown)


nils-a:
@ap0llo sorry for the late reply.
The change looks good.

As to the tests: There are currently no tests for styling of the `TextPrompt`. I would place the new tests in `./test/Spectre.Console.Tests/Unit/Prompts/TextPromptTests.cs`

ap0llo:
@nils-a I added some simple unit tests to ensure the styles are applied

nils-a (approved):
LGTM!

nils-a:
@ap0llo your changes have been merged, thanks for your contribution :+1:

PR #715, opened 2022-02-02
Custom stack trace and method name resolver for Exception formatting

phil-scott-78:
Newer versions of .NET clean up the exception messages internally for things like async and state machines. They clean the stack trace and also tweak the names. This gives a pretty big difference between our output and the default `ToString`. 

This allows the user to do their own tweaking to the method names and stack frames before the formatting. Or even better, use something like `Ben.Demystifer` to do the heavy lifting for them.

Here's the sample output using Ben.Demystifier. 

image

I'm adding

```csharp
AnsiConsole.WriteException(ex, new ExceptionSettings()
{
    StackTraceConverter = trace => new EnhancedStackTrace(trace),
    MethodNameResolver = frame => ((EnhancedStackFrame)frame).MethodInfo.Name
});
```

I'm adding two new properties - one that allows the stack trace to be messed with prior to formatting, and another that allows a custom method name resolver. If neither are set it falls back to the previous settings. Feedback please on the names or if y'all can think of a better way to hook in without needing to deal with a big mess of inheritance.

### DIFF
--- a/examples/Console/Exceptions/Exceptions.csproj
+++ b/examples/Console/Exceptions/Exceptions.csproj
@@ -12,4 +12,8 @@
     <ProjectReference Include="..\..\Shared\Shared.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Ben.Demystifier" Version="0.4.1" />
+  </ItemGroup>
+
 </Project>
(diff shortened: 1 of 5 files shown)


phil-scott-78:
Btw, I thought about looking at the code for the new `StackTrace` output and just making it "the way" for the default exception writer but I got intimidated pretty quickly

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/StackTrace.cs#L208

phil-scott-78:
The more I think about it the more I just want to clean the current output...

phil-scott-78:
I am going to clean the current output :-). gonna close this and get something new going.

PR #718, opened 2022-02-03
Improves exception rendering for async methods in .NET 6

phil-scott-78:
.NET 6 includes a new attribute that makes it easy to filter out stack frames we don't want to include. We can also improve upon the name of the method displayed when working with async calls.

## Before
before

## After
after

## Fallback
Previous versions of .NET without the attribute still work and have better method names with async calls, but the additional stack frames are still included
image


0xced:
This is going to conflict hard with #695. 😦

phil-scott-78:
Totally missed that PR. I suspect it wouldn't be terribly difficult for me to redo things to make it work on top of it, but I'd have to move a lot of the processing logic a lot closer to the rendering if I don't have the extra classes to smuggle the parsed data around.

Then again, it's not like we have a ton of use for the parsed data elsewhere so that might actually end up being a slightly cleaner solution altogether

phil-scott-78:
Ok, I looked at @0xced PR and building this on top of it I think it's the way to go. 

@patriksvensson and @nils-a if you all don't have any objections, I'm going to merge it and redo this PR on top of it.

patriksvensson:
@phil-scott-78 No objections from me!

phil-scott-78:
Great. I've merged the other PR and I'll redo this one!

Thanks for the heads up @0xced

phil-scott-78:
ok, redid this on top of @0xced work. Thanks again for the heads up. I quite like how they both came together.

patriksvensson (approved):
LGTM!

PR #719, opened 2022-02-03
Adding better type names for return types and parameters

phil-scott-78:
Uses the typenamehelper from Ben.Demystifer to help break down things like generic lists into their actual type display name for return types of methods and their paramters.

## Before
image

## After
image


phil-scott-78:
Added some code to break down params that are tuples. 

If someone really wanted to get after it, constructors, lambdas and local functions are in Ben.Demystify but the code for grabbing those is a bit more complex than looking at the types. But for this PR I think it's complete and ready for review.

patriksvensson (approved):
LGTM! 👍

PR #728, opened 2022-02-14
Docs redesign

phil-scott-78:
This has the whole redesign and completely lacks the reason I started it in the first place - adding in the documentation snippets. It might take a bit to get that nailed down because it'll require some reworking of the examples to make them a bit more documentation ready. A lot of good in what's already here so no sense waiting on that I suppose.


patriksvensson (approved):
A lot have changed, but build is passing so 👍

PR #732, opened 2022-02-16
Fixes #587 Support negative numbers as command option values

### LINKED ISSUE #587: Cannot specify negative number as command option value
As of version 0.40.0, it doesn't appear possible to specify a negative number as a command option.

Assuming I define a command that takes a decimal value as a command option, and that I want to pass a negative number value, neither of the following works:
```
mycommand -a -1.50
mycommand -a "-1.50"
```

Both attempts fail with this error `Not a valid name for a short option`. It appears that the command option parsing logic thinks that the -1.50 is supposed to be a command option.

I would expect that the quotes at least would effectively escape the numeric value. Is this a bug, or am I missing something?

Thanks.

jouniheikniemi:
Decided to create some unit tests for the tokenizer, as it would've been too much of a mess to test all of this through CommandApp. That's obviously up for debate, but at least the tokenizer tests helped me with this.

### DIFF
--- a/src/Spectre.Console/Cli/Internal/Parsing/CommandTreeTokenizer.cs
+++ b/src/Spectre.Console/Cli/Internal/Parsing/CommandTreeTokenizer.cs
@@ -235,13 +235,34 @@ internal static class CommandTreeTokenizer
                     ? new CommandTreeToken(CommandTreeToken.Kind.ShortOption, position, value, $"-{value}")
                     : new CommandTreeToken(CommandTreeToken.Kind.ShortOption, position + result.Count, value, value));
             }
+            else if (result.Count == 0 && char.IsDigit(current))
+            {
+                /* We require short options to be named with letters. Short options that start with a number
+                 * ("-1", "-2ab", "-3..7") may actually mean values (either for options or arguments) and will
+                 * be tokenized as strings. This block handles parsing those cases, but we only allow this
+                 * when the digit is the first character in the token (i.e. "-a1" is always an error), hence the
+                 * result.Count == 0 check above.
+                 */
+                string value = string.Empty;
+                while (!reader.ReachedEnd)
+                {
+                    char c = reader.Peek();
+
+                    if (char.IsWhiteSpace(c))
+                    {
+                        break;
+                    }
+
+                    value += c.ToString(CultureInfo.InvariantCulture);
+                    reader.Read();
+                }
+
+                value = "-" + value; // Prefix with the minus sign that we originally thought to mean a short option
+                result.Add(new CommandTreeToken(CommandTreeToken.Kind.String, position, value, value));
+            }
             else
             {
-                // Create a token representing the short option.
-                var tokenPosition = position + 1 + result.Count;
-                var represntation = current.ToString(CultureInfo.InvariantCulture);
-                var token = new CommandTreeToken(CommandTreeToken.Kind.ShortOption, tokenPosition, represntation, represntation);
-                throw CommandParseException.InvalidShortOptionName(reader.Original, token);
+                ThrowInvalidShortOptionName(current.ToString(CultureInfo.InvariantCulture));
             }
         }
 
@@ -254,6 +275,13 @@ internal static class CommandTreeTokenizer
         }
 
         return result;
+
+        void ThrowInvalidShortOptionName(string representation)
+        {
+            var tokenPosition = position + 1 + result.Count;
+            var token = new CommandTreeToken(CommandTreeToken.Kind.ShortOption, tokenPosition, representation, representation);
+            throw CommandParseException.InvalidShortOptionName(reader.Original, token);
+        }
     }
 
     private static CommandTreeToken ScanLongOption(CommandTreeTokenizerContext context, TextBuffer reader, int position)
--- a/test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
+++ b/test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
@@ -77,7 +77,7 @@ public static class CommandTreeTokenizerTests
         [Fact]
         public void With_Negative_Numeric_Prefixed_String_Value()
         {
-            var result = CommandTreeTokenizer.Tokenize(new [] { "-6..2 "});
+            var result = CommandTreeTokenizer.Tokenize(new[] { "-6..2 " });
             Assert.Empty(result.Remaining);
             var t = Assert.Single(result.Tokens);
             t.TokenKind.ShouldBe(CommandTreeToken.Kind.String);
--- a/test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
+++ b/test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
@@ -1,0 +1,90 @@
+namespace Spectre.Console.Tests.Unit.Cli;
+
+public static class CommandTreeTokenizerTests
+{
+    public sealed class ShortOptions
+    {
+        [Fact]
+        public void Valueless()
+        {
+            var result = CommandTreeTokenizer.Tokenize(new[] { "-a" });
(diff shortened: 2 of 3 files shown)


phil-scott-78 (commented):
Had some time to look through this. At a glance it looks good. But think we could get some tests that cover the invalid cases as described in the comments of the tokenizer for invalid items?

jouniheikniemi:
Hey @phil-scott-78, thanks for reviewing. Actually, CommandTreeTokenizer does have (logical) 100 % test coverage already. It is not immediately obvious, as this PR is the first stage that actually adds tests specifically for the Tokenizer; previously, the tokenizer has only been tested through CommandAppTests which test the entire parsing and execution pipeline. However, some of the test scenarios there (say, CommandAppTests.Parsing.InvalidShortOptionName.Should_Return_Correct_Text) do exercise these failure modes.

I would personally prefer to have 100 % coverage specifically with the tokenizer test suite, but writing all that is slightly beyond the scope of this issue. Would you expect to see (tokenizer level) tests for the failure scenarios of the ScanShortOptions tokenizer branch or what? TBH, the tokenizer test set in this PR doesn't cover all short option tokenization features (even the happy paths) either, grouped properties probably being the most important missing part. Given that there's no precedent for tokenizer tests in the project, I'm happy to do extra work to set a good example, let's just agree on the goal first.

src/Spectre.Console/Cli/Internal/Parsing/CommandTreeTokenizer.cs
```diff
@@ -235,13 +235,34 @@ private static IEnumerable<CommandTreeToken> ScanShortOptions(CommandTreeTokeniz
...
+                result.Add(new CommandTreeToken(CommandTreeToken.Kind.String, position, value, value));
+            }
             else
             {
-                // Create a token representing the short option.
-                var tokenPosition = position + 1 + result.Count;
-                var represntation = current.ToString(CultureInfo.InvariantCulture);
-                var token = new CommandTreeToken(CommandTreeToken.Kind.ShortOption, tokenPosition, represntation, represntation);
-                throw CommandParseException.InvalidShortOptionName(reader.Original, token);
+                ThrowInvalidShortOptionName(current.ToString(CultureInfo.InvariantCulture));
```
FrankRay78, line 265: I do not see the benefit of putting the local three-line logic into a separate private method, which also still references `reader.Original` from the calling function without passing this in as an explicit argument.  I suggest this private method be removed and the logic relocated inline here, instead.

test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
```diff
@@ -0,0 +1,90 @@
+namespace Spectre.Console.Tests.Unit.Cli;
+
+public static class CommandTreeTokenizerTests
```
FrankRay78, line 3: Fantastic that you have introduced explicit unit tests for the tokenizer! A positive move away from the general approach of 'unit tests' being located at the CommandApp level (nb. more like integration tests really).

Does this class need to be static?  Unit test classes aren't generally made static - suggest this is also removed to ensure consistency with the rest of the codebase.

test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            t.Position.ShouldBe(0);
+            t.Value.ShouldBe("a");
+            t.Representation.ShouldBe("-a");
+        }
+
+        [Theory]
+        [InlineData("-a:foo")]
+        [InlineData("-a=foo")]
+        public void With_Value(string param)
+        {
```
FrankRay78, line 25: // When

test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            var result = CommandTreeTokenizer.Tokenize(new[] { param });
+            Assert.Empty(result.Remaining);
+            Assert.Equal(2, result.Tokens.Count);
+
```
FrankRay78, line 29: // Then

test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            t.Value.ShouldBe("foo");
+            t.Representation.ShouldBe("foo");
+        }
+
+        [Theory]
+        [InlineData("-a:-1.5", null)]
+        [InlineData("-a=-1.5", null)]
+        [InlineData("-a", "-1.5")]
+        public void With_Negative_Numeric_Value(string firstArg, string secondArg)
+        {
```
FrankRay78, line 50: // Given

test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            List<string> args = new List<string>();
+            args.Add(firstArg);
+            if (secondArg != null)
+            {
+                args.Add(secondArg);
+            }
+
```
FrankRay78, line 57: // When

test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            var result = CommandTreeTokenizer.Tokenize(args);
+            Assert.Empty(result.Remaining);
+            Assert.Equal(2, result.Tokens.Count);
+
```
FrankRay78, line 61: // Then

test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            t.TokenKind.ShouldBe(CommandTreeToken.Kind.String);
+            t.IsGrouped.ShouldBe(false);
+            t.Position.ShouldBe(3);
+            t.Value.ShouldBe("-1.5");
+            t.Representation.ShouldBe("-1.5");
+        }
+
+        [Fact]
+        public void With_Negative_Numeric_Prefixed_String_Value()
+        {
```
FrankRay78, line 79: // When

test/Spectre.Console.Tests/Unit/Cli/CommandTreeTokenizerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            var result = CommandTreeTokenizer.Tokenize(new[] { "-6..2 " });
+            Assert.Empty(result.Remaining);
```
FrankRay78, line 81: // Then

FrankRay78 (changes requested):
Thank you for submitting this PR, my review comments are largely stylistic to ensure the changes fit stylistically with the rest of the codebase.

FrankRay78:
> this PR is the first stage that actually adds tests specifically for the Tokenizer; previously, the tokenizer has only been tested through CommandAppTests which test the entire parsing and execution pipeline

This is a significant improvement that @jouniheikniemi has introduced and I would be keen to see this PR, once review comments are addressed 😊, merged into the codebase.  Particularly since I have a raft of CommandTreeTokenizer tests I would like to add myself, ideally into the same class being introduced in this PR.

FrankRay78:
This should be closed as it's been fully included in https://github.com/spectreconsole/spectre.console/pull/1048  [I don't have permission to close another person's PR]

jouniheikniemi:
I was about to get back to your review comments, but it seems you'd rather pick up the ball yourself; fine with me. Closing the PR. Thanks!

FrankRay78:
I was keen to merge your excellent work @jouniheikniemi - thanks for your contribution.

PR #746, opened 2022-02-21
Add support for styling and aligning paths

patriksvensson:


phil-scott-78 (commented):
Everything looks good, but I noticed the example project is in the `Live` namespace rather than the `Paths` namespace. 

While not a big deal for the PR, it does make the docs get confused when looking for unique symbols throughout the solution.

patriksvensson:
@phil-scott-78 Good catch. I've updated the PR.

phil-scott-78 (approved):

PR #747, opened 2022-02-22
Optimizing build

phil-scott-78:
* Upgrades Statiq.Web to latest, which contains analyzer perf improvements
* Dropping script front matter when not needed from blog archives. 
* Trimming down the number of files tailwind examines when building the css file to only a subset of the docs. Tailwind looks at the content of all the files to determine which CSS classes to trim down to. We only need to look at a subset of the API to determine that, so this looks at a smaller namespace. 
* Updates Playwright to latest version
* Updates tailwind and asciienema to latest
* Removes npm packages that are no longer required

Locally you might have to do a `dotnet clean` because the protocol that Playwright uses has changed and you'll need to remove its cache. 

Also, with any luck removing the superfluous node packages and updating the others will autoclose the dependabot PRs

### DIFF
--- a/docs/Docs.csproj
+++ b/docs/Docs.csproj
@@ -33,13 +33,13 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Playwright" Version="1.13.0-next-1" />
+    <PackageReference Include="Microsoft.Playwright" Version="1.19.0" />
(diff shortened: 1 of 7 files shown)


patriksvensson (approved):

melvinmezzy (commented):
What’s this for and okay

PR #763, opened 2022-03-05
Adding a short template file for new documentation. 

phil-scott-78:
Adding a short template file people could use when documenting a new widget (or updating an existing one)


patriksvensson (approved):
Looks good to me!

PR #781, opened 2022-03-31
Change section heading

drewnoakes:
This appears to have been copy/pasted from progress.md, which has a similar structure.


nils-a (approved):
LGTM! Good Catch.

PR #824, opened 2022-05-04
Detect non-interactive console via Console.IsInputRedirected instead of Environment.UserInteractive

bastianeicher:
`Environment.UserInteractive` is `false` for SSH sessions on Windows, even though they are interactive. Detecting non-interactive sessions via redirected stdin instead seems more robust.

### DIFF
--- a/src/Spectre.Console/AnsiConsoleFactory.cs
+++ b/src/Spectre.Console/AnsiConsoleFactory.cs
@@ -40,7 +40,7 @@ public sealed class AnsiConsoleFactory
         var interactive = settings.Interactive == InteractionSupport.Yes;
         if (settings.Interactive == InteractionSupport.Detect)
         {
-            interactive = Environment.UserInteractive;
+            interactive = !System.Console.IsInputRedirected;
         }
 
         var profile = new Profile(output, encoding);


patriksvensson (approved):

patriksvensson:
@bastianeicher Could you rebase this PR against the `main` branch?

patriksvensson:
@bastianeicher The code in this PR does not compile. 

```
The type or namespace name 'IsInputRedirected' does not exist in the namespace 'Spectre.Console'
```

patriksvensson:
I was thinking, perhaps it would be better adding an SSH enriched, and checking whether or not `SSH_CLIENT` or `SSH_TTY` environment variables are defined.

bastianeicher:
> @bastianeicher The code in this PR does not compile.

Sorry about that. It did compile before the rebase but I didn't check it after. Fixed by replacing `Console` with `System.Console`.

> I was thinking, perhaps it would be better adding an SSH enriched, and checking whether or not `SSH_CLIENT`, `SSH_CONNECTION` or `SSH_TTY` environment variables are defined.

SSH on Windows is not the only case where `Environment.UserInteractive` doesn't return a reliable value. For example, on non-Windows platforms the value is simply [hardcoded to `true`](https://github.com/dotnet/runtime/blob/6cc686f5800feffb3f04bf48487cdfbd7b48a32d/src/libraries/System.Private.CoreLib/src/System/Environment.UnixOrBrowser.cs#L15). So I still think checking `System.Console.IsInputRedirected` here is a better option.

PR #837, opened 2022-05-24
Fix the "Escaping Interpolated Strings" documentation

0xced:
* `MarkupLineInterpolated` instead of `MarkupInterpolated`
* Use an actual interpolated string in sample code to make it valid


patriksvensson (approved):

patriksvensson:

PR #850, opened 2022-06-25
Check for null in TextPrompt

danielchalmers:
Should be slightly easier to debug.

```csharp
// Would throw "System.NullReferenceException : Object reference not set to an instance of an object."
AnsiConsole.Ask<string>(null);
// Now throws  "System.ArgumentNullException : Value cannot be null. (Parameter 'prompt')"
```

Matches ConfirmationPrompt:

https://github.com/spectreconsole/spectre.console/blob/564c7b8f4e44d4f83b89f1fdd1edf44da1d91da1/src/Spectre.Console/Prompts/ConfirmationPrompt.cs#L48


patriksvensson (approved):

patriksvensson:

PR #617, opened 2021-11-08
Add a prompt that allows a selection of a row in a table

FantasyTeddy:
This PR adds a very basic version of a prompt that displays a table with a highlighted row representing the selection.

PR #888, opened 2022-07-02
Add a prompt that allows a selection of a row in a table

FantasyTeddy:
This PR adds a very basic version of a prompt that displays a table with a highlighted row representing the selection.

This is a copy of #617.


SadE54:
You mean , that could be used as a selector for autocompleting ?

FantasyTeddy:

PR #892, opened 2022-07-05
feat: optional cancellationtoken for asynchronous commands

PsychoNineSix:
Our internal tooling using spectre has commands that can run quite long, and we execute the commandapp within an input loop. Ctrl+c currently closes the application and we have to restart it in 'interactive mode'
Our preference would be to make ctrl+c cancel the command when running by passing a cancellationtoken.

### DIFF
--- a/src/Spectre.Console.Cli/CommandApp.cs
+++ b/src/Spectre.Console.Cli/CommandApp.cs
@@ -1,3 +1,5 @@
+using System.Threading;
+
 namespace Spectre.Console.Cli;
 
 /// <summary>
@@ -52,7 +54,7 @@ public sealed class CommandApp : ICommandApp
     /// <returns>The exit code from the executed command.</returns>
     public int Run(IEnumerable<string> args)
     {
-        return RunAsync(args).GetAwaiter().GetResult();
+        return RunAsync(args, null).GetAwaiter().GetResult();
     }
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


PsychoNineSix:
Not sure if adding this to the context is the best way, but creating overloads for this seemed overkill.

src/Spectre.Console.Cli/CommandApp.cs
```diff
@@ -52,7 +54,7 @@ public void SetDefaultCommand<TCommand>()
     /// <returns>The exit code from the executed command.</returns>
     public int Run(IEnumerable<string> args)
     {
-        return RunAsync(args).GetAwaiter().GetResult();
+        return RunAsync(args, null).GetAwaiter().GetResult();
```
PsychoNineSix, line 57:
```suggestion
        return RunAsync(args).GetAwaiter().GetResult();
```

src/Spectre.Console.Cli/CommandApp.cs
```diff
@@ -61,6 +63,17 @@ public int Run(IEnumerable<string> args)
     /// <param name="args">The arguments.</param>
     /// <returns>The exit code from the executed command.</returns>
     public async Task<int> RunAsync(IEnumerable<string> args)
+    {
+        return await RunAsync(args, null);
```
PsychoNineSix, line 67:
```suggestion
        return await RunAsync(args);
```

src/Spectre.Console.Cli/CommandApp.cs
```diff
@@ -61,6 +63,17 @@ public int Run(IEnumerable<string> args)
...
+        return await RunAsync(args);
+    }
+
+    /// <summary>
+    /// Runs the command line application with specified arguments.
+    /// </summary>
+    /// <param name="args">The arguments.</param>
+    /// <param name="token">Optional cancellation token.</param>
+    /// <returns>The exit code from the executed command.</returns>
+    public async Task<int> RunAsync(IEnumerable<string> args, CancellationToken? token = null)
```
PsychoNineSix, line 76: My question around here is what would you prefer:
- overload without argument (with `RunAsync(args, null);`
- optional parameter (like it is now, and remove the overload)

PsychoNineSix:
> @PsychoNineSix Any specific reason why you closed this?

Wasn't sure about the implementation and time was an issue to look into it, might still do so when i get more time on my hands though!

PR #911, opened 2022-08-02
Modified MarkupTokenizer, so that escaped markup inside markup is valid

### LINKED ISSUE #903: EscapeMarkup doesn't work when using [link=...]
**Information**
 - OS: [Windows]
 - Version: [0.44.0]
 - Terminal: [Windows Terminal]

**Describe the bug**
Escape doesn't work in some cases. It seems in combination with [link] it throws an exception `Encountered unescaped ']' token at position 56`

**To Reproduce**
```csharp
            var path = "C:\\Dev\\ExampleProjects\\TestApp\\TestApp[-].UnitTests\\StrykerOutput\\2022-07-22.15-27-40\\reports\\mutation-report.html";
            // the next line prints our path
            AnsiConsole.Console.MarkupLineInterpolated($"[Green]{path}[/]");
            // the next line throws an exception
            AnsiConsole.Console.MarkupLineInterpolated($"[Green][link={path}]{path}[/][/]");
```
**Expected behavior**
No exception.

**Additional context**
This is an actual use case in Stryker.

nils-a:
fixes #903

### DIFF
--- a/src/Spectre.Console/Internal/Text/Markup/MarkupTokenizer.cs
+++ b/src/Spectre.Console/Internal/Text/Markup/MarkupTokenizer.cs
@@ -24,134 +24,154 @@ internal sealed class MarkupTokenizer : IDisposable
         }
 
         var current = _reader.Peek();
-        if (current == '[')
+        return current == '[' ? ReadMarkup() : ReadText();
+    }
+
+    private bool ReadText()
+    {
+        var position = _reader.Position;
+        var builder = new StringBuilder();
+
+        var encounteredClosing = false;
+        while (!_reader.Eof)
         {
-            var position = _reader.Position;
-
-            _reader.Read();
-
-            if (_reader.Eof)
-            {
-                throw new InvalidOperationException($"Encountered malformed markup tag at position {_reader.Position}.");
-            }
-
-            current = _reader.Peek();
+            var current = _reader.Peek();
             if (current == '[')
             {
+                // markup encountered. Stop processing.
+                break;
+            }
+
+            // If we find a closing tag (']') there must be two of them.
+            if (current == ']')
+            {
+                if (encounteredClosing)
+                {
+                    _reader.Read();
+                    encounteredClosing = false;
+                    continue;
+                }
+
+                encounteredClosing = true;
+            }
+            else
+            {
+                if (encounteredClosing)
+                {
+                    throw new InvalidOperationException(
+                        $"Encountered unescaped ']' token at position {_reader.Position}.");
+                }
+            }
+
+            builder.Append(_reader.Read());
+        }
+
+        if (encounteredClosing)
+        {
+            throw new InvalidOperationException($"Encountered unescaped ']' token at position {_reader.Position}.");
+        }
+
+        Current = new MarkupToken(MarkupTokenKind.Text, builder.ToString(), position);
+        return true;
+    }
+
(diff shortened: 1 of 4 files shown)


phil-scott-78 (approved):
Been a minute since I've looked in here, but things look good to me. 

Maybe it has been my time away from digging into the guts, but if you have to jump back in to fix the formatting issue it might be nice to comment out the difference scenarios we are checking for in the tokenizer. The nesting is getting kind of deep and it was tough for my admittedly sleep deprived brain to follow :-)

nils-a:
Yes - it was kind of late yesterday, but I wanted to get the code out, so you could pick on it 😁 
I'm not really fond of the nesting either - I had a hard time grasping what was going on. Then again, I was never good at writing parsers. 🤷

src/Spectre.Console/Internal/Text/Markup/MarkupTokenizer.cs
```diff
@@ -113,7 +138,7 @@ public bool MoveNext()
                     if (encounteredClosing)
                     {
                         throw new InvalidOperationException(
-                            $"Encountered unescaped ']' token at position {_reader.Position}");
+                            $"Encountered unescaped ']' token at position {_reader.Position}.");
```
patriksvensson, line 141: I understand the tempation to add the dot here, but we should remove it so we're consistent with the rest of the library (or change all of them) 😅

src/Spectre.Console/Internal/Text/Markup/MarkupTokenizer.cs
```diff
@@ -66,21 +66,46 @@ public bool MoveNext()
...
+                }
+
+                if (next == '[')
+                {
+                    _reader.Read();
+                }
+                else
+                {
+                    throw new InvalidOperationException(
+                        $"Encountered malformed markup tag at position {_reader.Position - 1}.");
```
patriksvensson, line 100: See comment about punctuation.

src/Spectre.Console/Internal/Text/Markup/MarkupTokenizer.cs
```diff
@@ -122,7 +147,7 @@ public bool MoveNext()
 
             if (encounteredClosing)
             {
-                throw new InvalidOperationException($"Encountered unescaped ']' token at position {_reader.Position}");
+                throw new InvalidOperationException($"Encountered unescaped ']' token at position {_reader.Position}.");
```
patriksvensson, line 150: See comment about punctuation.

patriksvensson (changes requested):

src/Spectre.Console/Internal/Text/Markup/MarkupTokenizer.cs
nils-a: I've reverted that change. I started to make them consistent but stopped - that is a far too big change to slide it in here 🤣

nils-a:
I've restructured the `MarkupTokenizer` a bit so that it's hopefully a bit more readable/understandable.

I let the text of the exceptions basically as they were before this PR - I'll start a PR to make the exceptions all have the "same" punctuation as soon as this one's fine.

rouke-broersma:
Thank you for fixing this! Looking forward to a release :)

patriksvensson (approved):

PR #924, opened 2022-08-17
(#916) fixed missing call to Validate when using CommandConstructorBinder

### LINKED ISSUE #916: Validate is not called if a CommandSettings class defines a constructor
**Information**
 - OS: Windows 11 Pro
 - Version: 0.44.0
 - Terminal: Windows Terminal / PowerShell 7

**Describe the bug**
Spectre CLI doesn't seem to call `Validate` if a constructor is provided for a `CommandSettings` class.

**To Reproduce**

```c#
namespace TestingSpectre;

using System.Diagnostics.CodeAnalysis;
using Spectre.Console;
using Spectre.Console.Cli;

internal class Program
{
    public class DogCommand : Command<DogCommand.DogSettings>
    {
        public sealed class DogSettings : CommandSettings
        {
            public DogSettings(string name)
            {
                Name = name;
            }

            [CommandOption("-n|--name <VALUE>")]
            public string Name { get; }

            public override ValidationResult Validate()
            {
                if (Name == "Tiger")
                {
                    return ValidationResult.Error("Tiger is not a dog name!");
                }

                return ValidationResult.Success();
            }
        }

        public override int Execute([NotNull] CommandContext context, [NotNull] DogSettings settings)
        {
            AnsiConsole.WriteLine($"The dog's name is {settings.Name}");
            return 0;
        }
    }

    static int Main()
    {
        CommandApp app = new();
        app.Configure(config =>
        {
            config.AddCommand<DogCommand>("dog");
        });
        return app.Run(new[] { "dog", "-n", "Tiger" });
    }
}
```

**Expected behavior**
In the example above, invocation when calling `Run` should fail with an error but it runs without error.  When removing the constructor and setting the `Name` property to `{ get; set; }`, the validator is called as expected.

Using constructors for such classes is important when nullable is enabled or the compiler will throw a warning stating that all non-nullable properties must be set by the constructor of the class.

**Screenshots**
N/A

**Additional context**
N/A

Thanks heaps in advance!
Fotis

nils-a:
fixes #916

### DIFF
--- a/test/Spectre.Console.Cli.Tests/Data/Settings/TurtleSettings.cs
+++ b/test/Spectre.Console.Cli.Tests/Data/Settings/TurtleSettings.cs
@@ -1,0 +1,16 @@
+namespace Spectre.Console.Tests.Data;
+
+public sealed class TurtleSettings : ReptileSettings
+{
+    public TurtleSettings(string name)
+    {
+        Name = name;
+    }
+
+    public override ValidationResult Validate()
+    {
+        return Name != "Lonely George"
+            ? ValidationResult.Error("Only 'Lonely George' is valid name for a turtle!")
+            : ValidationResult.Success();
+    }
+}
--- a/src/Spectre.Console.Cli/Internal/Binding/CommandConstructorBinder.cs
+++ b/src/Spectre.Console.Cli/Internal/Binding/CommandConstructorBinder.cs
@@ -45,6 +45,13 @@ internal static class CommandConstructorBinder
             }
         }
 
+        // Validate the settings.
+        var validationResult = settings.Validate();
+        if (!validationResult.Successful)
+        {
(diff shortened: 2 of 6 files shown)


test/Spectre.Console.Cli.Tests/Data/Settings/TurtleSettings.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public TurtleSettings(string name)
+    {
+        Name = name;
+    }
+
+    public override ValidationResult Validate()
+    {
+        return Name != "Lonely George"
+            ? ValidationResult.Error("Only 'Lonely George' is valid name for a turtle!")
```
patriksvensson, line 13: Love this 😄

patriksvensson (approved):

PR #972, opened 2022-09-21
Fix @ being used in figlet font

Saalvage:
Currently, @ being used in a figlet font causes it to not load, as *all* @ are removed from every line, which causes line length differences and is incorrect.

According to http://www.jave.de/figlet/figfont.html section "FIGcharacter Data", (which, I'm not sure is equivalent to a standard specification?) theoretically:
> The FIGdriver will eliminate the last block of consecutive equal characters from each line of sub-characters when the font is read in.

.. and @ and $ are only conventions for the chosen terminating character.

However, considering @ is currently hardcoded as the only valid terminating character and the fact that it has not caused any problems thus far it is safe to assume that it is fine in practice to stick by this convention.

-------------------
It might be worth considering additional fonts for the tests to potentially uncover similar issues, but that is beyond the scope of this pull request.


patriksvensson (approved):

patriksvensson:

PR #976, opened 2022-09-24
Allow selections to wrap around

Saalvage:
It honestly surprised me this wasn't standard behavior.

Tests aren't exhaustive but considering there weren't any tests for the state before this should be good enough.

### DIFF
--- a/src/Spectre.Console/Prompts/List/ListPrompt.cs
+++ b/src/Spectre.Console/Prompts/List/ListPrompt.cs
@@ -15,7 +15,8 @@ internal sealed class ListPrompt<T>
     public async Task<ListPromptState<T>> Show(
         ListPromptTree<T> tree,
         CancellationToken cancellationToken,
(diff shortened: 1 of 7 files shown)


phil-scott-78:
Besides the issue with the code style, the code itself looks good. But to be honest I'm not 100% sure what I'm seeing when it comes to what this is trying to solve.

This is the behavior I'm seeing with the current release. There something I'm missing? 

Animation

To be clear, I'm not saying there isn't room for this to be better. Just trying to track down current behavior vs new based on the code

phil-scott-78:
Oh jeez, my toddlers killed more brain cells than I thought yesterday. It's so obvious this morning 🤣

I'll take a look again tonight and unless @patriksvensson or @nils-a have any differing opinion on the property name I'll get it merged

nils-a:
`WrapAround` is fine for me. :+1:

phil-scott-78 (approved):

PR #996, opened 2022-10-05
Allow letter keys (A-Z) to cycle indexes in Prompts

tznind:
Adds the ability to press letter keys (A-Z) to cycle selection to next item beginning with that letter in Prompt.

I have made changes only to `ListPromptState.cs` as this is where Up and Down are being handled.  Let me know if there is a better place for it or if this isn't a feature you want to incorporate.

loopWithLetters
_Check up/down works then pressing 'M' repeatedly then 'L' then End followed by clycling 'O'_

### DIFF
--- a/src/Spectre.Console/Prompts/List/ListPromptState.cs
+++ b/src/Spectre.Console/Prompts/List/ListPromptState.cs
@@ -21,15 +21,49 @@ internal sealed class ListPromptState<T>
 
     public bool Update(ConsoleKey key)
     {
-        var index = key switch
+        var index = Index;
+
+        // if user presses a letter key
+        if (key >= ConsoleKey.A && key <= ConsoleKey.Z)
         {
-            ConsoleKey.UpArrow => Index - 1,
-            ConsoleKey.DownArrow => Index + 1,
+            var keyStruck = new string((char)key, 1);
+
+            // find indexes of items that start with the letter
+            int[] indexes = Items.Select((item, idx) => (item, idx))
+                               .Where(k => k.item.Data.ToString()?.StartsWith(keyStruck, StringComparison.InvariantCultureIgnoreCase) ?? false)
+                               .Select(k => k.idx)
+                               .ToArray();
+
+            // if there are items begining with this letter
+            if (indexes.Length > 0)
+            {
+                // is one of them currently selected?
+                var currentlySelected = Array.IndexOf(indexes, index);
+
+                if (currentlySelected == -1)
+                {
+                    // we are not currently selecting any item begining with the struck key
+                    // so jump to first item in list that begins with the letter
+                    index = indexes[0];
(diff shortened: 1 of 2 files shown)


src/Spectre.Console/Prompts/List/ListPromptState.cs
```diff
@@ -21,15 +21,48 @@ public ListPromptState(IReadOnlyList<ListPromptItem<T>> items, int pageSize, boo
...
+        var index = Index;
+
+        // if user presses a letter key
+        if (key >= ConsoleKey.A && key <= ConsoleKey.Z)
+        {
+            var keyStruck = new string((char)key, 1);
+
+            // find indexes of items that start with the letter
+            int[] matchingIndexes = Items.Select((item, idx) => (item, idx))
+                                   .Where(k => k.item.Data.ToString()?.StartsWith(keyStruck, StringComparison.InvariantCultureIgnoreCase) ?? false)
```
thoemmi, line 33: `k.item.Data` must not be `string` but can be any type of object. You have to respect a potential convert. See how the data is rendered e.g. in `SelectionPrompt`: https://github.com/spectreconsole/spectre.console/blob/8fa53f8fb19fbb4fb6e330d8d127519c4a0b19b3/src/Spectre.Console/Prompts/SelectionPrompt.cs#L166

When you want to jump to an item with the specific starting character, you have to look for the string that is actually displayed.

tznind: Thanks for pointing this out, I had missed that.  I have added the `converter` arg to `ListPromptState<T>.Update` method so it can check the actual string contents.  I added it as a default parameter to the method but if you would rather I create a new overload instead then I can update to that approach.

I had to add the converter property to `IListPromptStrategy<T>` since that is what `ListPrompt.cs` has access to.  And `ListPrompt` is the one calling `state.Update`.

The only alternative I could think of was moving the A-Z logic out of the state class and into the `HandleInput` methods instead.  But this would be replicated between `MultiSelectPrompt` and `SelectionPrompt` so I think a bad alternative.

thoemmi (approved):

patriksvensson:
Hi! 

Thanks so much for taking the time to submit this PR. 
I really appreciate the contribution!

Unfortunately, I haven’t had the capacity to review this in a timely manner, and since it was opened, the repository has changed significantly. Because of that, this PR is no longer in a state where it can be reviewed or merged as-is. This is not your fault, but mine, and I'm sorry about this.

I’m going to close it for now to help keep things tidy. That said, if you’re still interested in contributing, please feel free to open a new pull request that’s up to date with the current state of the repository, and I’ll be happy to take a look when I can. *I'll promise I'll do better this time around.*

Thanks again for your work and for your interest in the project!

PR #1001, opened 2022-10-06
Add in return for Abort on MultiSelectionPrompt

### LINKED ISSUE #851: Esc Key to exit selection

BenjaminMichaelis:
Fixes #851

### DIFF
--- a/src/Spectre.Console/Prompts/List/ListPrompt.cs
+++ b/src/Spectre.Console/Prompts/List/ListPrompt.cs
@@ -16,7 +16,8 @@ internal sealed class ListPrompt<T>
         ListPromptTree<T> tree,
         CancellationToken cancellationToken,
         int requestedPageSize = 15,
-        bool wrapAround = false)
+        bool wrapAround = false,
+        bool allowAbort = true)
     {
         if (tree is null)
         {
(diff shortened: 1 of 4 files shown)


BenjaminMichaelis:
@patriksvensson To follow up with my question on this issue, is this all you are looking for a return of the Result? Or also some handling of some sort?

patriksvensson:
@BenjaminMichaelis 

We call `HandleInput` from here: https://github.com/spectreconsole/spectre.console/blob/main/src/Spectre.Console/Prompts/List/ListPrompt.cs#L59

1. We would need to add a check here for the `Abort` result, and in that case, break.
2. We should probably add a new `required` boolean to the `ListPrompt.Show` method, and before aborting, check if we're allowed to, so we don't abort if the prompts are expected to produce some kind of result.

What do you think?

BenjaminMichaelis:
I think that makes sense, Let me check that I fully understand what you mean by trying to implement it

BenjaminMichaelis:
How does this look? @patriksvensson

patriksvensson:
@BenjaminMichaelis I think it looks good! Would it be possible to add a test as well?

patriksvensson:
The build problems should have been resolved now.
Could you rebase your PR branch against the main branch?

BenjaminMichaelis:
Just rebased @patriksvensson, added in a small test, but as I am still getting used to this library, not positive if there are better things I should be testing for/or how to since this feels minor. I'm not positive how to get the ListPromptInputResult out from the test view since we return a state

patriksvensson:
@BenjaminMichaelis Your test actually showed that there is a flaw in the design.
We always return the selected item(s) and in this case, we want to inform the caller that the operation was aborted.

@nils-a @phil-scott-78 Any suggestions on how we could solve this in an extendable way?

BenjaminMichaelis:
Ahh, to make sure I am following, so the state index should still be 0 in fact like the test shows but in addition be letting the caller know that it didn't move because this was an aborted case?

patriksvensson:
@BenjaminMichaelis Correct, but I think we need to discuss how this should be done. Would like some input from the other maintainers about this.

patriksvensson:
@BenjaminMichaelis I'm sorry, but turns out there already is an active pull request for this item: https://github.com/spectreconsole/spectre.console/pull/711

I'm sorry I didn't realize this before, I could have saved you from wasting your time 😞

BenjaminMichaelis:
> @BenjaminMichaelis I'm sorry, but turns out there already is an active pull request for this item: #711
> 
> I'm sorry I didn't realize this before, I could have saved you from wasting your time 😞

Aw bummer, I'll keep looking for more to do :)

Kamakazi152:
@patriksvensson Am I looking at this right and seeing #711 was never finished? If so, could @BenjaminMichaelis submit his changes again? This feature is still not implemented, and it looks like Benjamin already got this figured out.

PR #1005, opened 2022-10-12
Allow `PACKET` key on MultiSelectionPrompt

nilaoda:
When use mobile RD Client, can not toggle items through `<SPACE>`, it actually `<PACKET>` key. (I tested iOS, not sure 
 about other platforms)

Here is my test:

SCTest

To toggle items, must use Screen Keyboard...

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,5 @@
 {
-  "projects": [ "src", "tests" ],
   "sdk": {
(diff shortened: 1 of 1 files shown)


nilaoda:
Sorry, i closed this PR due to a mistake...

PR #1027, opened 2022-10-23
modified tokenizer to not break on on ]]] at the end of a style

nils-a:
@patriksvensson, @phil-scott-78 can't say I like this change, but it gets the job done for the moment.

There are two things I don't like in particular:
- This is the first time the tokenizer gains knowledge about styles and how they work. Not sure if that's a good thing.
- Running `StringBuilder.ToString()` in a loop feels weird. Not sure if it would easier/better/more performant to drop the `StringBuilder` here and simply concatenate the chars together.

I'd appreciate any input.


patriksvensson (approved):

patriksvensson:
Merged! 👍

PR #1028, opened 2022-10-24
Make word wrapping consider initial paragraph spacing as a margin

lonely-generation:
Hey there!

What I'm trying to achieve is to keep paragraph indentation, even when word wrapping is applied.

So, using the following code:

```c#
using Spectre.Console;

var tab = new string(' ', 4);

var oneTab = tab + """
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec ultrices lorem sit amet quam malesuada, quis cursus lacus posuere. Aenean id felis a lorem molestie eleifend. Mauris vehicula convallis sagittis.
    """;

var twoTabs = tab + tab + """
    Donec ut rutrum quam, ut pretium magna. Nunc lacinia in enim sit amet semper. Phasellus sed ante id nibh tincidunt pulvinar id vel erat. Mauris elementum ut orci at blandit. Praesent in dictum magna. Donec tincidunt ornare enim. 
    """;

AnsiConsole.MarkupLine(oneTab);
AnsiConsole.MarkupLine(twoTabs);
AnsiConsole.MarkupLine(oneTab);
```

I'd like to have this (PR): 
BI6DSS92c3

But not this (current implementation):
1Apgjk9lVK

So this is what this PR is about. 
It'll break default behavior, so some Console flags might be added for such feature.

### DIFF
--- a/src/Spectre.Console/Rendering/SegmentLineIterator.cs
+++ b/src/Spectre.Console/Rendering/SegmentLineIterator.cs
@@ -15,6 +15,11 @@ public sealed class SegmentLineIterator : IEnumerator<Segment>
     /// </summary>
     public Segment Current { get; private set; }
 
+    /// <summary>
+    /// Gets the current line.
+    /// </summary>
+    public SegmentLine? CurrentLine { get; private set; }
+
     /// <inheritdoc/>
     object? IEnumerator.Current => Current;
 
@@ -34,6 +39,7 @@ public sealed class SegmentLineIterator : IEnumerator<Segment>
         _lines = new List<SegmentLine>(lines);
 
         Current = Segment.Empty;
+        CurrentLine = null;
     }
 
     /// <inheritdoc/>
@@ -100,6 +106,7 @@ public sealed class SegmentLineIterator : IEnumerator<Segment>
         // Reset the flag
         _lineBreakEmitted = false;
 
+        CurrentLine = _lines[_currentLine];
         Current = _lines[_currentLine][_currentIndex];
         return true;
     }
--- a/src/Spectre.Console/Widgets/Paragraph.cs
+++ b/src/Spectre.Console/Widgets/Paragraph.cs
@@ -197,6 +197,8 @@ public sealed class Paragraph : Renderable, IAlignable, IOverflowable
 
         var lines = new List<SegmentLine>();
         var line = new SegmentLine();
+        var lineSeed = (Segment?)null;
+        var currentLine = (SegmentLine?)null;
 
         var newLine = true;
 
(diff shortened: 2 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/spectreconsole/spectre.console?pullRequest=1028) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/spectreconsole/spectre.console?pullRequest=1028) before we can accept your contribution.<br/><br/>:x: lonely-generation [sign now](https://cla.dotnetfoundation.org/spectreconsole/spectre.console?pullRequest=1028)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/spectreconsole/spectre.console?pullRequest=1028) it.</sub>

patriksvensson (changes requested):
@lonely-generation Thanks for your PR!

I think this feature would need to be opt-in to avoid breaking people who rely on the previous behavior. Could you also add some tests that prove the new behavior acts like it's supposed to?

lonely-generation:
To enable this feature one should use `AnsiConsole.Profile.PreserveSpacing = true;` or the same for any console object correspondingly

patriksvensson:
@lonely-generation I've been thinking, and I'm a bit hesitant about controlling this kind of behavior with a global flag.
The effect you're trying to achieve could be done using a padding widget.

```csharp
AnsiConsole.Write(new Padder(new Markup("Hello world").PadLeft(4));
```

lonely-generation:
@patriksvensson Thank you! This is exactly what I've been looking for!
Somehow I couldn't figure this way since I though this is impossible to use nested `IRenderable`, especially in cases when I have no idea about how deep it can go (not a case for a Table). 

For anyone who's wondering how nesting can be done, check out the code below. This is just an example, of course those nested statements are hard to read and should be refactored.

```c#
using Spectre.Console;

var one = "The dotnet command has two functions:";
var two = "It provides commands for working with .NET projects. For example, dotnet build builds a project. Each command defines its own options and arguments. All commands support the --help option for printing out brief documentation about how to use the command.";

AnsiConsole.MarkupLine("[bold]Description[/]");
AnsiConsole.Write(new Padder(new Markup(one)).Padding(4, 0, 0, 0));
AnsiConsole.Write(new Padder(new Padder(new Markup(two)).Padding(4, 0, 0, 0)).Padding(4, 0, 0, 0));

// Same but using Panel instead
AnsiConsole.MarkupLine("[bold]Description[/]");
AnsiConsole.Write(new Panel(new Markup(one)).NoBorder().Padding(4, 0, 0, 0));
AnsiConsole.Write(new Panel(new Panel(new Markup(two)).NoBorder().Padding(4, 0, 0, 0)).NoBorder().Padding(4, 0, 0, 0));
```

P1kAlMiG2K

Guess this PR no longer makes sense and can be closed. Sorry and thanks again for your time

PR #1029, opened 2022-10-25
GH-193: support options before arguments

FrankRay78:
Allow options to be specified on the command line ahead of arguments without the `Flags cannot be assigned a value.` error occurring.

This PR should fully satisfy the following issue: https://github.com/spectreconsole/spectre.console/issues/193 (and its open duplicate, https://github.com/spectreconsole/spectre.console/issues/825)

Extensive unit tests have been authored (largely in `CommandAppTests, Spectre.Console.Tests.Unit.Cli`), covering the impact of this change and ensuring good functioning.

Issue #959, opened 2022-09-10
POSIX: Invalid handling of parameter values starting with double quotes (`"`), e.g. `--order-by='"quoted"` value'


sandreas:
Important: This is a follow-up issue (#891), since it was closed without possibility to give feedback, reopen or answering my follow-up questions. I really think this is clearly a bug of `spectre` and it should be further investigated, because input values are *MODIFIED* without any developer interaction... It may also be possible, that this leads to a *SECURITY FLAW* because parameters are mistreated, but I'm no expert.

This also may not be reproducible in any OS (e.g. on Windows) AND also not via *Debugger* or *IDE*, since command line parameter handling is different. To reproduce the issue, I recommend using Linux and compile the given test program in release mode.

The last answer I got stated this about argument parsing:

> Unfortunately, this is out of our hands. We are just parsing the args that .NET gets from its command line handler. In fact, on Windows I get different results from you with different examples.

I think, this is wrong, because at least something in `spectre` seems to `Trim` whitespaces (Details: https://github.com/spectreconsole/spectre.console/issues/891#issuecomment-1179605527) beause the following example does not modify parameters, while `spectre` does.

```c#
// Program.cs, dotnet 6
    Console.WriteLine("== Environment.CommandLine ==");
    Console.WriteLine(Environment.CommandLine);
    Console.WriteLine("");

    Console.WriteLine("==  Environment.GetCommandLineArgs() ==");
    foreach(var a in Environment.GetCommandLineArgs()){
        Console.WriteLine(a);
    }
    Console.WriteLine("");

    Console.WriteLine("== args ==");
    foreach(var b in args){
        Console.WriteLine(b);
    }
    return 0;
```

**Information**
 - OS: Linux, Ubuntu 18.04 LTS
 - Version: 0.45
 - Terminal: xterm-256color, zsh

**Describe the bug**
Parameter values cannot start with double quotes (`"`). They are either replaced or handled completely wrong. Additionally, parameter values are TRIMMED (remove leading spaces), so adding a space to workaround this is not possible.  

The following example even cuts off the parameter value and treats everything after the space as EXTRA argument, in the following case `value` is handled as positional argument and not part of `--order-by`:
```bash
cli-tester test --order-by '"quoted" value' input.mp3
```
This may be a security flaw under specific circumstances.

More examples of invalid handling:
```bash
# `quoted` instead of `"quoted" value` and cut off handling as positional argument
cli-tester test --order-by '"quoted" value' input.mp3
[...]

Issue #890, opened 2022-07-03
Value starting with `-` is mistreated as extra option 


sandreas:
**Information**
 - OS: Linux, Ubuntu 18.04 LTS
 - Version: 0.44
 - Terminal: xterm-256color, zsh

**Describe the bug**
Providing a valid value starting with `-` throws an exception / error, when using the following syntax:
```
# value `-size` is short form for `size descending`
cli-tester test input.mp3 --order-by "-size"

Error: Option 'order-by' is defined but no value has been provided.

       test input.mp3 --order-by -size
                      ^^^^^^^^^^ No value provided
```

Using the syntax with `=` as value separator works like expected:

```
cli-tester test input.mp3 --order-by="-size"

# works like expected, prints out the unmodified option value for `--order-by`
order-by=<-size>

```

**To Reproduce**

- Compile https://github.com/sandreas/tone/tree/main/cli-tester
- Run the above examples

**Expected behavior**
I would expect that option valid values with leading `-` would be correctly handled when NOT using `=` as separator, otherwise the behaviour is inconsistent in my opinion.

FrankRay78:
This is a value bug imho - see the justification here: https://github.com/spectreconsole/spectre.console/issues/959#issuecomment-1292741847

FrankRay78:
This issue can be closed/marked as complete @patriksvensson, now that PR https://github.com/spectreconsole/spectre.console/pull/1048 has been successfully merged.

Issue #842, opened 2022-06-05
Wrong handling of empty string and space values (e.g. `--order-by=""` or `--order-by=" "`)


sandreas:
**Information**
 - OS: Linux, Ubuntu 18.04 LTS
 - Version: 0.44
 - Terminal: xterm-256color, zsh
 
**Describe the bug**
Providing an empty value for a string option is not possible in my test project. 

```
cli-tester test input.mp3 --order-by=""


Error: Expected an option value.

       test input.mp3 --order-by=
                                 ^ Did you forget the option value?

```

I would consider an empty string as a valid value. Even stranger, if I provide a space (` `) as value, it gets trimmed, although I never did that myself:

```
cli-tester test input.mp3 --order-by=" "
order-by=<> # no error, but no space!!?
```

**To Reproduce**

- Compile https://github.com/sandreas/tone/tree/main/cli-tester
- Run the above examples

**Expected behavior**
- If I provide an empty option (`cli-tester test input.mp3 --order-by=""`), the value should be considered as valid
- If I provide a space as option (`cli-tester test input.mp3 --order-by=" "`), the given value should be preserved as is and not trimmed (I would expect that for ANY value)

Maybe I am doing something wrong?!

FrankRay78:
This is a value bug imho - see the justification here: https://github.com/spectreconsole/spectre.console/issues/959#issuecomment-1292741847

FrankRay78:
This issue can be closed/marked as complete @patriksvensson, now that PR https://github.com/spectreconsole/spectre.console/pull/1048 has been successfully merged.

PR #1036, opened 2022-10-31
GH-959: Invalid handling of parameter values with quotes (double and single, enclosed and open), whitespace, hyphens and special characters (eg. emojis)

FrankRay78:
This PR fully implements https://github.com/spectreconsole/spectre.console/issues/959, https://github.com/spectreconsole/spectre.console/issues/842, https://github.com/spectreconsole/spectre.console/issues/890, namely POSIX compliant handling of quotes (double and single, enclosed and open), whitespace, hyphens and special characters (eg. emojis) when arriving in the `IEnumerable<string> args` array, agnostic of the syntax required by the surrounding shell to do so.

A significant increase to unit test coverage accompanies this PR, particularly focused on the `CommandTreeTokenizer` class. 41 new unit tests cover the parsing and tokenisation of string, long and short options. Minimal changes to existing tests were necessary, and only when the existing implementation incorrectly (according to POSIX) handled quotes/whitespace/hyphens etc.

When this PR is considered collectively with https://github.com/spectreconsole/spectre.console/pull/1029 and https://github.com/spectreconsole/spectre.console/pull/732, these PR's would make a significant enhancement to command line parsing, extending behaviour across all platforms (behaviour already familiar to Mac and Linux users), and IMHO, would make an excellent minor release.


test/Spectre.Console.Cli.Tests/Unit/CommandAppTests.Parsing.cs
```diff
@@ -356,29 +356,6 @@ public Task Should_Return_Correct_Text_For_Short_Option_If_Strict_Mode_Is_Enable
             }
         }
 
-        [UsesVerify]
-        [ExpectationPath("UnterminatedQuote")]
-        public sealed class UnterminatedQuote
-        {
-            [Fact]
-            [Expectation("Test_1")]
-            public Task Should_Return_Correct_Text()
```
patriksvensson, line 365: Would prefer if we kept this test to ensure backwards compatibility.

test/Spectre.Console.Cli.Tests/Unit/CommandAppTests.Parsing.cs
```diff
@@ -609,24 +586,6 @@ public void Should_Allow_Special_Symbols_In_Name(string option)
                 result.ShouldBe("Error: Command 'dog' is missing required argument 'AGE'.");
             }
         }
-
-        [Fact]
-        [Expectation("Quoted_Strings")]
-        public Task Should_Parse_Quoted_Strings_Correctly()
```
patriksvensson, line 615: Would prefer if we kept this test to ensure backwards compatibility.

FrankRay78: I can add it back, that's not a problem.  

However, let me explain why I re/moved this first, see if that makes a difference...

The same [functionally equivalent] test is now located in CommandAppTests+Remaining (Spectre.Console.Tests.Unit.Cli), namely ` public void Should_Preserve_Quotes_Hyphen_Delimiters()`:

```
            var result = app.Run(new[]
            {
                "animal", "4", "dog", "12", "--",
                "/c", "\"set && pause\"",
                "Name=\" -Rufus --' ",
            });

            // Then
            result.Context.Remaining.Raw.Count.ShouldBe(3);
            result.Context.Remaining.Raw[0].ShouldBe("/c");
            result.Context.Remaining.Raw[1].ShouldBe("\"set && pause\"");
```

I moved it here as every other test in `test/Spectre.Console.Cli.Tests/Unit/CommandAppTests.Parsing.cs` is actually handling error states, not happy paths.  Plus it seemed an appropriate new location alongside the other dedicated remaining args unit tests.

Also, and perhaps most importantly, by re/moving the `Should_Parse_Quoted_Strings_Correctly()` test out of `CommandAppTests.Parsing`, I felt it paved the way to rename the whole class (perhaps more appropriately) to `CommandAppTests.ParsingErrors`. 

[nb. now that I've written the above, perhaps the renaming should happen as I suggest, but with `CommandAppTests.Parsing` also being resurrected with a single method, the `public Task Should_Parse_Quoted_Strings_Correctly()` to ensure backwards compatibility as you suggest]

Happy to do any of the above, if you let me know which one.  _(sorry if I'm making things overly complicated...)_

test/Spectre.Console.Cli.Tests/Unit/CommandAppTests.Parsing.cs
FrankRay78: The unit test, `UnterminatedQuote.Should_Return_Correct_Text()` throws an error when a terminating quote isn't found when parsing a command line argument and converting it into (one or more) tokens. 

I deleted the test because `CommandTreeTokenizer` now handles both terminated and unterminated quotes when received in an argument.

`CommandTreeTokenizerTests` evidences this behaviour, see the various tests with the following decorations:

```
        // Double-quote handling
        [InlineData("\"")]
        [InlineData("\"\"")]
        [InlineData("\"Rufus\"")]
        [InlineData("\" Rufus\"")]
        [InlineData("\"-R\"")]
        [InlineData("\"-Rufus\"")]
        [InlineData("\" -Rufus\"")]

        // Single-quote handling
        [InlineData("'")]
        [InlineData("''")]
        [InlineData("'Rufus'")]
        [InlineData("' Rufus'")]
        [InlineData("'-R'")]
        [InlineData("'-Rufus'")]
        [InlineData("' -Rufus'")]

        [InlineData(" \" -Rufus -- ")]
        [InlineData("Name=\" -Rufus --' ")]
```

[nb. `ScanQuotedString(...)` was also removed in this PR as the ScanString can now handle quotes without special consideration]

FrankRay78:
Closing as superseded by https://github.com/spectreconsole/spectre.console/pull/1048

PR #1041, opened 2022-11-01
Add Layout widget

patriksvensson:
There are a lot of things that need to be fixed so we can have a proper layout widget.
Don't worry, we're not turning Spectre.Console to a TUI framework. It's just useful in some scenarios.


src/Spectre.Console/Widgets/Layout/Layout.cs
```diff
@@ -0,0 +1,277 @@
...
+    public Layout this[string name]
+    {
+        get => GetLayout(name);
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="Layout"/> class.
+    /// </summary>
+    /// <param name="name">The layout name.</param>
+    public Layout(string? name = null)
```
phil-scott-78, line 109: Would an optional `IRenderable` make sense here for the default content too? Then if that's set `.SplitRows` would throw perhaps?

patriksvensson: Good idea. Will add a new constructor.

src/Spectre.Console/Widgets/Layout/Layout.cs
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    /// <summary>
+    /// Updates the containing <see cref="IRenderable"/>.
+    /// </summary>
+    /// <param name="renderable">The renderable to use for this layout.</param>
+    /// /// <returns>The same instance so that multiple calls can be chained.</returns>
+    public Layout Update(IRenderable renderable)
+    {
+        _renderable = renderable ?? new LayoutPlaceholder(this);
```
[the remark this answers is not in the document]
patriksvensson: I don't think splitting should throw. You could theoretically hide all child layouts and let it fall back to the parent one.

nils-a (approved):
LGTM!
(Except the failing Build 😁 )

PR #1045, opened 2022-11-03
Issue 644

alex-eshock:
Fixes #644 where settings containing arrays needed to be documented.

Issue #653, opened 2021-12-04
CaseSensitivity should be documented somewhere


nils-a:
As it affects `Commands` and long `Options` it should probably be in https://spectreconsole.net/cli/commandapp and https://spectreconsole.net/cli/settings

alex-eshock:
I can work on this

alex-eshock:
I've submitted a pull request for the issue now.

PR #1049, opened 2022-11-04
Added Test Cases For Asynchronous Methods/Commands to Resolve Issue 676

### LINKED ISSUE #676: Some unit tests other examples are needed for AsyncCommand.
I was chatting with @patriksvensson on Twitter and he suggest I look at AsyncCommand for an example project I posted on youtube. 

I checked it out, but I see no unit tests for AsynCommand or examples. 

I suggest that be rectified.

Cheers!

alex-eshock:
Fixes #676 where there were little to no test cases that tested asynchronicity within the application.

I've implemented 2 test classes that have multiple test cases that are similar to `CommandAppTest` classes to provide a more tangible example between asynchronous and synchronous runtime.
- `AsyncCommandTest.cs`
- `AsyncCommandTests.Exceptions.cs`

@wguner also implemented a class that serves as the base for testing asynchronous commands called `AsyncTester.cs`.


docs/input/cli/commandApp.md
```diff
@@ -21,6 +21,10 @@ app.Configure(config =>
 });
 ```
 
+## Case Sensitivity
+
+Commands are in `PascalCase`, also known as `UpperCamelCase`.
```
patriksvensson, line 26: Are these changes related to #653?

I think this doesn't seem right. Commands that we provide to an application such `myexe.com command --option 1` are case sensitive in the sense that `myexe.com Command --option 1` (notice the capital 'C') would return an error if the command was named `command`. It's not the `Command` class that is case sensitive.

docs/input/cli/settings.md
```diff
@@ -24,6 +24,10 @@ public sealed class MyCommandSettings : CommandSettings
 
 This setting file tells `Spectre.Console.Cli` that our command has two parameters. One is marked as a `CommandArgument`, the other is a `CommandOption`.
 
+## Case Sensitivity
+
+CommandsSettings are in `PascalCase`, also known as `UpperCamelCase`.
```
patriksvensson, line 29: Are these changes related to #653?

I think this doesn't seem right. Commands that we provide to an application such `myexe.com command --option 1` are case sensitive in the sense that `myexe.com Command --option 1` (notice the capital 'C') would return an error if the command was named `command`. It's not the `Command` class that is case sensitive.

docs/input/cli/settings.md
```diff
@@ -48,6 +52,8 @@ public string? LastName { get; set; }
 * Options with a single character must be preceded by a single dash (e.g. `-c`).
 * Multi-character options must be preceded by two dashes (e.g. `--count`).
 
+`NOTE:` There can only be one `argument vector`, and it needs to be the last argument of the command.
```
patriksvensson, line 55: @phil-scott-78 Is this correct?

src/Spectre.Console.Testing/Cli/AsyncTester.cs
```diff
@@ -0,0 +1,133 @@
+namespace Spectre.Console.Testing;
+
+/// <summary>
+/// A <see cref="CommandApp"/> test harness.
+/// </summary>
+public class AsyncTester
```
patriksvensson, line 6: I don't think we need a whole new class for this. The old `CommandAppTester` could be augmented with a `RunAsync` method instead.

patriksvensson (changes requested):

docs/input/cli/settings.md
patriksvensson: I think this belongs to PR #1045 and has accidentally been included.

docs/input/cli/commandApp.md
alex-eshock: I just removed the documentation i accidentally added. Things should be good from here.

docs/input/cli/settings.md
alex-eshock: I just removed the documentation i accidentally added. Things should be good from here.

docs/input/cli/settings.md
alex-eshock: I just removed the documentation i accidentally added. Things should be good from here.

src/Spectre.Console.Testing/Cli/AsyncTester.cs
alex-eshock: I moved some of the class contents to `CommandAppTester` and made a `RunAsync` method within that class. I have removed the `AsyncTester.cs` class as well.

FrankRay78:
I'm closing this PR given it's age, the unresolved review comments and the underlying progress of the main branch. Thank you @alex-eshock for your efforts and feel free to contribute to spectre.console again in the future.

Issue #1059, opened 2022-11-09
Ability to add command line flags and their values to remaining arguments if they cannot be set on the settings class


FrankRay78:
**Is your feature request related to a problem? Please describe.**
There is an open issue in the cake project, https://github.com/cake-build/cake/issues/3279, which has been triaged and accepted as an improvement, to 'Allow users to use any of the Cake runner arguments as build script arguments'

An example of what this practically means is as follows. Here is a flag currently defined on the cake `DefaultCommandSettings` class:

```
        [CommandOption("--version|--ver")]
        [Description("Displays version information.")]
        public bool ShowVersion { get; set; }
```

And normally the user would invoke this by `dotnet cake --version`.  

However, sometimes the user would like to specify a version number and _have it pass_ to their build script instead. Currently they must write `dotnet cake -- --version=1.2.3` to explicitly ensure the version number is parsed as a remaining argument, not a command setting.

The above issue, https://github.com/cake-build/cake/issues/3279, would allow `dotnet cake --version=1.2.3` to add version=1.2.3 to the remaining arguments list, and `dotnet cake --version` to set the ShowVersion property to true (ie. continue operating as current).

**Describe the solution you'd like**
A configuration setting that informs the command parsing logic to catch any exceptions thrown when setting values on the command settings class (eg. `Flags cannot be assigned a value.`) and instead add the option and its value to the remaining arguments list. This configuration setting could be turned off as a default to ensure backward compatibility with existing behaviour.

_(This solution depends on **https://github.com/spectreconsole/spectre.console/pull/1048** being merged, so that command options (and their values) work nicely ahead of any arguments)_

**Describe alternatives you've considered**
Should this feature be deemed as not viable for spectre.console for any reason, I would look to see how the exception could be caught higher up in the cake codebase, and/or some kind of extension point being opened within the command parsing logic perhaps. And then subsequently some way to inject the caught value back into the remaining args from the cake codebase [pretty ugly sounding]. I think, however, the solution proposed above is a clean and generic improvement to spectre console whilst also preserving existing behaviour.

nb. please assign the issue to me @patriksvensson if it successfully makes it out of triage 🙂

PR #1048, opened 2022-11-03
Command line argument parsing improvements

FrankRay78:
The following three PR's, when taken jointly, represent a significant enhancement to existing command line parsing behaviour: https://github.com/spectreconsole/spectre.console/pull/1036, https://github.com/spectreconsole/spectre.console/pull/1029, https://github.com/spectreconsole/spectre.console/pull/1040

However, they all make changes to CommandTreeTokenizer and introduce their own unit tests specifically for the CommandTreeTokenizer, and as such, merge conflicts will arise once any one of the PR's has been merged.

This PR encompasses all three PR's, successfully merged and with all conflicts resolved. The respective git histories have also been squashed so the changes for each issue can be more clearly examined. 

All issues addressed by this single combined PR are: https://github.com/spectreconsole/spectre.console/issues/959, https://github.com/spectreconsole/spectre.console/issues/842, https://github.com/spectreconsole/spectre.console/issues/890, https://github.com/spectreconsole/spectre.console/issues/193 (and its open duplicate, https://github.com/spectreconsole/spectre.console/issues/825), https://github.com/spectreconsole/spectre.console/issues/587

As it's quite a big change, I'm happy to help in any way required @patriksvensson for you to accept in full, in part, with or without amendment etc.

PR #1060, opened 2022-11-11
1059 Ability to add command line values to remaining arguments if they cannot be set on the settings class

FrankRay78:
This PR implements #1059, 'Ability to add command line flags and their values to remaining arguments if they cannot be set on the settings class'

A configuration setting called `ConvertFlagsToRemainingArgumentsIfCannotBeAssigned` has been added to the CommandAppSettings interface/class, which determines whether this behaviour has been turned on (defaults to off to ensure backward compatibility). CommandTreeParser checks this setting prior to deciding whether to throw the `Flags cannot be set.` exception, or alternatively, to add the flag and its value to the remaining arguments collection.

Whilst this PR consists of 5 commits, it's the last commit titled `Allow flags to be converted to remaining arguments if they cannot be assigned` (11 Nov 2022) that contains the changes for this PR - the previous commits are because I branched from https://github.com/spectreconsole/spectre.console/pull/1048, which is a dependency because command options (and their values) should work nicely ahead of any arguments.

**Important: https://github.com/spectreconsole/spectre.console/pull/1048 should be merged ahead of this PR**


FrankRay78:
Closing as I have replaced this PR with https://github.com/spectreconsole/spectre.console/pull/1102

PR #1077, opened 2022-11-17
Proper CancellationToken handling for ListPrompt. Fixes #1076.

### LINKED ISSUE #1076: Prompt Cancellation is broken
**Information**
 - OS: Windows
 - Version: 0.45.0
 - Terminal: Windows Terminal

**Describe the bug**
Issue #417 mentions that cancellation should be supported, and PR #477 is said to fix it, but it you, for example, press Ctrl+C while selecting items from a list, the process gets stuck. 

I debugged the source, and I believe I found the root cause.
This is the only line of code that checks for cancellation, at all, and this is correct:
https://github.com/spectreconsole/spectre.console/blob/52c1d9122bce4c73ae9205e2f2aff01cba03a4c4/src/Spectre.Console/Internal/DefaultInput.cs#L41-L44

The problem is that the SelectionPrompt, and likely all other Prompts, are doing this:
https://github.com/spectreconsole/spectre.console/blob/a9b97fa4dc0e868e1835b91ceed3641085d282ae/src/Spectre.Console/Prompts/List/ListPrompt.cs#L49-L54

They should **_also_** check for the cancellation of the cancellation token before they call the underlying method.

**To Reproduce**
```c#
await new SelectionPrompt<T>()
...
    .AddChoices(choices)
    .ShowAsync(AnsiConsole.Console, ct)
```
Then simply cancel the cancellation token `ct`, and the process will get stuck.

**Expected behavior**
The SelectionPrompt.ShowAsync(...) method should return when the cancellation token is cancelled.

azchohfi:
Proper CancellationToken handling for ListPrompt.

### DIFF
--- a/src/Spectre.Console/Prompts/List/ListPrompt.cs
+++ b/src/Spectre.Console/Prompts/List/ListPrompt.cs
@@ -48,6 +48,8 @@ internal sealed class ListPrompt<T>
 
             while (true)
             {
+                cancellationToken.ThrowIfCancellationRequested();
+
                 var rawKey = await _console.Input.ReadKeyAsync(true, cancellationToken).ConfigureAwait(false);
                 if (rawKey == null)
                 {


caesay:
This is definitely an issue in TextPrompt, possibly also MultiSelectionPrompt. Ideally a fix would cover all of the possible prompts.

PR #1091, opened 2022-11-25
Add possibility to set description and/or data for the default command

### LINKED ISSUE #247: Passing data for Default command
It looks like you can't set `Data` property for "default" command.

Either use `Data` defined in config for same type of command or ability to specify context for the "default" command, same as regular command.

**Additional context**

Example code:
```csharp
var externalDataObject = new { AdditionalContext = "myContext" };

var app = new CommandApp();
app.SetDefaultCommand<MyCommand>();
app.Configure(config =>
{
    config
        .AddCommand<MyCommand>("myCommand")
        .WithData(externalDataObject);

    config
        .AddCommand<MyAnotherCommand>("myAnotherCommand");
});
```
If you call the app like `myApp myCommand --myArg xxx` then the `MyCommand` will be called and "context" will have `externalDataObject` value, but if you call it like `myapp --myArg xxx`  then the `MyCommand` will be called but "context" will not have `externalDataObject` value.

0xced:
Sample usage:
```csharp
var app = new CommandApp<DefaultCommand>().WithDescription("Default command description").WithData(new object());
```

```csharp
var app = new CommandApp();
app.SetDefaultCommand<DefaultCommand>(description: "Default command description", data: new object());
```

Fixes #247


0xced:
@FrankRay78 Would you mind reviewing this pull request too? Next year is fine too. 😉

FrankRay78:
Yes of course @0xced - I've been quite sick lately, but once mended, will be back to my duties.

0xced:
Ouch, get well soon!

Gnbrkm41:
Just a note that examples (normally configurable via `WithExample` on commands) remain unconfigurable for default commands; Worth getting it in with the rest of the config as well?

0xced:
Actually, the `AddExample` method is exposed on `IConfigurator`, so you can already provide examples for the default command like this:

```csharp
var app = new CommandApp();
app.Configure(config =>
{
    config.AddExample(new[] { "file.txt", "--ignore-case" });
});
```

This is why I have limited this pull request to _data_ + _description_ only.

FrankRay78:
Hello @0xced, I'm looking at this now... 

I noticed that in `CommandApp` you chose to implement a number of overloaded methods:

```csharp
public void SetDefaultCommand<TCommand>(string description)
public void SetDefaultCommand<TCommand>(object data)
public void SetDefaultCommand<TCommand>(string description, object data)
```

Rather than follow the convention used in `CommandAppOfT`, namely the ability to chain together calls to the setters in a fluent syntax eg:

```csharp
public CommandApp WithDescription(string description)
public CommandApp WithData(object data)
```

Unless there was a good reason for the difference in approach, I'd prefer to standardise across `CommandApp` and  `CommandAppOfT` and adopted the approach taken in `CommandAppOfT`.

src/Spectre.Console.Cli/CommandAppOfT.cs
```diff
@@ -8,6 +8,7 @@ public sealed class CommandApp<TDefaultCommand> : ICommandApp
     where TDefaultCommand : class, ICommand
 {
     private readonly CommandApp _app;
+    private readonly ConfiguredCommand _defaultCommand;
```
FrankRay78, line 11: An alternative would be:

```csharp
private ConfiguredCommand? DefaultCommand => _app.GetConfigurator().DefaultCommand;
```

src/Spectre.Console.Cli/CommandAppOfT.cs
```diff
@@ -47,4 +48,26 @@ public Task<int> RunAsync(IEnumerable<string> args)
     {
         return _app.RunAsync(args);
     }
+
+    /// <summary>
+    /// Sets the description of the default command.
+    /// </summary>
+    /// <param name="description">The default command description.</param>
+    /// <returns>The same <see cref="CommandApp{TDefaultCommand}"/> instance so that multiple calls can be chained.</returns>
+    public CommandApp<TDefaultCommand> WithDescription(string description)
```
FrankRay78, line 57: An alternative would be:

```csharp
    public CommandApp<TDefaultCommand> WithDescription(string description)
    {
        if (DefaultCommand != null)
        {
            DefaultCommand.Description = description;
        }

        return this;
    }
```

Given that `CommandAppOfT` always has a default command, you could remove the null check and suppress the intellisense warning of a possible null.  

However, if you kept the null check in place, then this method could be copied to CommandApp almost verbatim (the return type needs to be changed to non-generic) for it's `WithDescription` method

src/Spectre.Console.Cli/Internal/Configuration/Configurator.cs
```diff
@@ -25,11 +25,12 @@ public void AddExample(string[] args)
         Examples.Add(args);
     }
 
-    public void SetDefaultCommand<TDefaultCommand>()
+    public ConfiguredCommand SetDefaultCommand<TDefaultCommand>()
```
FrankRay78, line 28: The alternatives offered above would remove the need to change this method/class.

FrankRay78:
I didn't mean to scare you off with my review comments @0xced...  I assume you are just enjoying the holiday period.  However, all of the above is open for discussion, as it's not a dictatorship 🙂

0xced:
No worries, I haven't been scared at all but I was indeed enjoying the holidays away from computer. 😉

I agree that the `SetDefaultCommand` methods with multiple overload was not a very good design. I initially did it in order not to break binary compatibility even though it's not strictly required by semantic versioning since Spectre.Console has not yet reached version 1.0.

I have introduced a new `DefaultCommandBuilder` class in 6e63da1ab1f4f1765b86d6139838004644487241 which will allow to chain the default command configuration like its generic `CommandApp<TDefaultCommand>` counterpart:

```csharp
var app = new CommandApp();
app.SetDefaultCommand<DefaultCommand>().WithDescription("Default command description").WithData(new object());
```

src/Spectre.Console.Cli/Internal/Configuration/Configurator.cs
0xced: The `Configurator` class is internal so changing a method to return something instead of `void` is a non issue in my opinion.

Also, this allows to exposes a non nullable default command which I considered very carefully during implementation to avoid having a nullable `private ConfiguredCommand? DefaultCommand` as you suggested above. Having a nullable field or property and suppress the nullability with a `!` later is something I'd prefer to avoid if possible.

FrankRay78:
Thanks @0xced , I'll make time this week to look at this PR again.... and hopefully get it merged.

FrankRay78:
Excellent work @0xced and I really like the fluent builder.  

I noticed that the builder was very similar to/a cut down version of `Spectre.Console.Cli.CommandConfigurator`, and so to ensure consistency, I've renamed `DefaultCommandBuilder` to `DefaultCommandConfigurator`, and moved the class alongside the `CommandConfigurator`.  I hope that makes sense.

FrankRay78 (approved):

Issue #702, opened 2022-01-24
Function to render help text


WajahatAliAbid:
**Is your feature request related to a problem? Please describe.**
I was looking to find a way to show command help during execution of a command. For example, I have main command of cli set to display fidglet text. But I want to show command help right after.   

**Describe the solution you'd like**
Ideally, having a function in AnsiConsole or CommandContext to render command help. For example
```csharp
public override int Execute(CommandContext context)
{
   AnsiConsole.WriteHelp();
}
```
I am not sure if there exists a function to provide such implementation. If there is, please refer me to that.

Simonl9l:
Would be good to add a:

```C#
public override int Execute(CommandContext context)
{
   AnsiConsole.WriteVersion();
}
```

rerasolutions:
Good question, is there a way to call the help? I use a default command and I would like to show the help there

CrossBound:
I am starting a greenfield application and I just learned today about Spectre.Console & Spectre.Console.Cli. It sounded great so I decided to use it. Very shortly after starting I realized when errors occur (such as user input invalid command, input, etc.) the error is displayed very nicely, but the help is not rendered. I think it's very common (and in my case desirable) to show the help when the user messes up the input. So it should show the error, then immediately write out the help to describe how to use the command. With the current implementation I don't see a way to accomplish that and it looks like this issue targets the problem. 

Unfortunately it appears this issue has been open a very long time and the pull request to resolve this has been open for a while as well. I was looking forward to using Spectre.Console.CLI but, unfortunately, I guess because of this issue I'll check if System.CommandLine will allow me to do this, and if so, I'll be switching to that framework instead.

FWIW: My use case is I would like to provide a custom exception handler. If the exception is due to a problem parsing the input, I would like to display the error and then write help to the console. To accomplish this, I would just need a way to access the help writer from within an exception handler.

CrossBound:
FYI: It appears System.CommandLine supports my use case by default with no extra configuration required.

FrankRay78:
@CrossBound I prepared the pull request to help address the issue, it's just waiting for review and could have been expedited if asked. It's a shame you didn't stick around for longer so that we might have been able to accommodate your use case more fully, however thank you for leaving feedback. Needless to say this is not a commercially oriented project and yet the judicious time and effort, delays included in that, does mean the codebase is a very high quality. We'd welcome pull requests if you did want to contribute in the future.

Issue #1093, opened 2022-11-26
Separate initialization from execution in `CommandApp`


rcdailey:
As a user of the API, I'd like the ability to perform global setup and teardown in my application without utilizing base classes for my `Command<>` / `AsyncCommand<>` subclasses.

## Is your feature request related to a problem? Please describe.

At the moment, if I want to perform some global initialization and/or cleanup, I am required to use base classes or other obscure hacks. The problem with the base class solution is that it creates a lot of constructor boilerplate and makes the code difficult to maintain when changes are needed in that common base class, especially when that change involves modifying dependencies (parameters for DI).

## Describe the solution you'd like

I'd like the logic in `CommandApp.RunAsync()` to be separated such that the following is separated into distinct phases of execution controlled by the user of the API:

1. Parsing of the command tree. Registration of internal types and dependencies. Finally, creation of the implementation for `ITypeResolver`.
2. Execution of the actual commands.

This allows me to do the following:

1. Perform setup of my command line application as normal.
2. Invoke a new method (or even better: Use the builder pattern) to initiate step 1 above (e.g. `CommandApp.Setup()`).
3. Access the `ITypeResolver` via some property (e.g. `CommandApp.Resolver`) to perform DI resolution for any dependencies needed for my global initialization/cleanup.
4. Perform global initialization as needed in `Program.Main()`
5. Invoke `CommandApp.Run()` and store the result.
6. Perform global cleanup as needed in `Program.Main()`
7. Return the result of the previously executed `CommandApp.Run()` command.

## Describe alternatives you've considered

1. Abusing my implementation of `ITypeRegistrar` to invoke a callback to perform custom initialization when `ITypeRegistrar.Build()` is called.
2. Implementing a base class shared by all command handlers that implements `Command<>` or `AsyncCommand<>`.

## Additional context

I am willing to implement the changes needed here. I've already played around with some ideas in a local branch and landed on two approaches:

1. Quick & dirty: Add new methods to `CommandApp` with the goal of reducing footprint. Not the most ideal change and am not completely happy with it, but it at least gets that separation I'm aiming for.
2. Implement something like `CommandAppBuilder` that uses a syntax similar to `new CommandAppBuilder().WithConfiguration(...).Build()` to construct a `CommandApp` and return it, which will be responsible for the first phase of initialization. The returned `CommandApp` would actually be composed in a wrapper object that only allows users to invoke `Run()` / `RunAsync()` on it.

[...]

PR #1133, opened 2023-01-13
Can render command help and application version from inside a command

FrankRay78:
Fully implements https://github.com/spectreconsole/spectre.console/issues/702

---

Also, starts to introduce a clean entry point for allowing spectre.console users (in the future) to override the default HelpWriter with their own implementation, here is where the DI will need to happen, `Spectre.Console.Cli.CommandExecutor` line 54: 

```csharp
var helpProvider = new HelpWriter(model, parsedResult, configuration.Settings.ShowOptionDefaultValues) as IHelpProvider;
```

### DIFF
--- a/src/Spectre.Console.Cli/CommandContext.cs
+++ b/src/Spectre.Console.Cli/CommandContext.cs
@@ -3,7 +3,7 @@ namespace Spectre.Console.Cli;
 /// <summary>
 /// Represents a command context.
 /// </summary>
-public sealed class CommandContext
+public sealed class CommandContext : IHelpProvider
 {
     /// <summary>
     /// Gets the remaining arguments.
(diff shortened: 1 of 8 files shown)


FrankRay78:

patriksvensson (commented):
I've been thinking about this for a while.

Wouldn't it be better if we exposed an interface `IHelpWriter` (like you have in this PR) that contains a method to create the help text from a simplified version of `CommandModel`?

```csharp
public interface IHelpProvider
{
    IEnumerable<IRenderable> GetHelp(SomeSimplifiedCommandModel model);
}
```

This way, people could design their help text however they want.

FrankRay78:
I've been thinking about this as well, and I agree @patriksvensson .   

It also starts to tie in with this https://github.com/spectreconsole/spectre.console/issues/1093, which is about separating out initialisation/allowing users to register their own concrete implementations in the DI factory ahead of command parsing etc.

I'll have a go at updating this PR and resubmit.

FrankRay78:
FYI. I'm going to start reworking this PR @patriksvensson.  Assuming the help writer improvements in https://github.com/spectreconsole/spectre.console/pull/1252 are generally on point, I'm going to branch from that as a better starting point than main.

FrankRay78:
Closing, as superseded by https://github.com/spectreconsole/spectre.console/pull/1259

PR #1151, opened 2023-02-04
feat: Allow case-insensitive confirmation prompt

### LINKED ISSUE #1119: Add support for case insensitive confirmation
image
"Y" or "N" are not valid answers. Not a big problem, but it would be better to answers be case insensitive.

MartinZikmund:
closes #1119

Adding this for https://github.com/unoplatform/uno.check/issues/121 😉

### DIFF
--- a/src/Spectre.Console/Prompts/ConfirmationPrompt.cs
+++ b/src/Spectre.Console/Prompts/ConfirmationPrompt.cs
@@ -40,6 +40,12 @@ public sealed class ConfirmationPrompt : IPrompt<bool>
     public bool ShowDefaultValue { get; set; } = true;
 
     /// <summary>
+    /// Gets or sets a value indicating whether the confirmation
+    /// should use case insensitive matching.
+    /// </summary>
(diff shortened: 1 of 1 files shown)


patriksvensson (approved):
LGTM!

MartinZikmund:
Wow, that was fast! Thanks @patriksvensson!

0xced:
How about exposing a `StringComparison` property directly (that would default to `StringComparison.CurrentCulture`) instead of a `bool` property? So that it would be possible for the user to choose `InvariantCultureIgnoreCase` or  `OrdinalIgnoreCase` too?

MartinZikmund:
That is a neat idea. It would be a breaking change, but as it was a recent addition, I guess only Uno.Check is currently using it, what do you think @patriksvensson ?

0xced:
Given it was introduced last week in a **preview** version of a library which has not reached a stable 1.0 version I think it should be fine. But the final decision is up to the maintainers of course.

patriksvensson:
@MartinZikmund @0xced Sounds like a good idea to me!

PR #1152, opened 2023-02-04
Ensure correct comparer is used for `TextPrompt`

MartinZikmund:
Follow up to #1151, to correctly capture the response, the correct comparer needs to be used for `TextPrompt` choices. Verified this now works as expected.


patriksvensson (approved):
LGTM!

PR #1160, opened 2023-02-11
Add an implicit operator to convert from Color to Style

0xced:
Often a Style is only composed of a foreground color. Having an implicit operator greatly reduces verbosity.

Before the implicit operator:
```csharp
AnsiConsole.Console.Write("All your base are belong to us", new Style(Color.Yellow));
```

After the implicit operator:
```csharp
AnsiConsole.Console.Write("All your base are belong to us", Color.Yellow);
```


FrankRay78 (approved):

0xced:
Thanks for merging!

FrankRay78:
Thank you @0xced, another excellent contribution greatly received.    You also taught me about implicit conversions with your neat code, I can't believe having used C# for 20 years I've never seen that trick before.   I'll get on to your other PR soon....

PR #1172, opened 2023-02-20
Do not register analyzer if SpectreConsole is not available in the current compilation

meziantou:
It's common to register analyzers in Directory.Build.props or Central packages. This means the analyzers are registered for all projects even if they don't contain a reference to Spectre.Console. This PR checks in CompilationStart if the types are actually available before registering analyzers.


phil-scott-78 (approved):
Thanks for the explanation on registering with the build.props. Makes sense.

PR #1174, opened 2023-02-21
Use StringComparison.Ordinal instead of culture-sensitive comparisons

meziantou:
`StartsWith` and `EndsWith` are culture sensitive. Use `StringComparison.Ordinal` to avoid potential problem because of the current culture.

### DIFF
--- a/src/Spectre.Console/Widgets/TextPath.cs
+++ b/src/Spectre.Console/Widgets/TextPath.cs
@@ -52,12 +52,12 @@ public sealed class TextPath : IRenderable, IHasJustification
         _parts = path.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
 
         // Rooted Unix path?
-        if (path.StartsWith("/"))
+        if (path.StartsWith("/", StringComparison.Ordinal))
         {
             _rooted = true;
             _parts = new[] { "/" }.Concat(_parts).ToArray();
         }
-        else if (_parts.Length > 0 && _parts[0].EndsWith(":"))
+        else if (_parts.Length > 0 && _parts[0].EndsWith(":", StringComparison.Ordinal))
         {
             // Rooted Windows path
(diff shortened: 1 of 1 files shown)


src/Spectre.Console/Widgets/TextPath.cs
```diff
@@ -52,12 +52,12 @@ public TextPath(string path)
         _parts = path.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
 
         // Rooted Unix path?
-        if (path.StartsWith("/"))
+        if (path.StartsWith("/", StringComparison.Ordinal))
```
gfoidl, line 55:
```suggestion
        if (path.StartsWith('/'))
```
use the char-overload?

src/Spectre.Console/Widgets/TextPath.cs
```diff
@@ -52,12 +52,12 @@ public TextPath(string path)
...
         {
             _rooted = true;
             _parts = new[] { "/" }.Concat(_parts).ToArray();
         }
-        else if (_parts.Length > 0 && _parts[0].EndsWith(":"))
+        else if (_parts.Length > 0 && _parts[0].EndsWith(":", StringComparison.Ordinal))
```
gfoidl, line 60:
```suggestion
        else if (_parts.Length > 0 && _parts[0].EndsWith(`:`))
```

src/Spectre.Console/Widgets/TextPath.cs
meziantou: This overload is not available in netstandard2.0. I can add an extension method in `StringExtensions.cs` if you want

````
#if NETSTANDARD2_0
internal static bool StartsWith(this str, char c) => str.Length > 0 && str[0] == 'c';
#endfif
````

patriksvensson (approved):
LGTM!

PR #1178, opened 2023-02-21
Added TextPath.NoLineBreak flag

### LINKED ISSUE #1177: Make line breaks optional for TextPath
Currently `TextPath` includes an unconditional line break after rendering:

```
public IEnumerable<Segment> Render(RenderOptions options, int maxWidth)
    ...

   // Insert a line break
    parts.Add(Segment.LineBreak);

    return parts;
}

```
It would be nice if this line break could be made optional. This will allow us to write addition text after the rendered "path".

My current use case for this is inside an image resizing app: I output the filename (using TextPath) before resizing it, and after processing I output the time taken. Using TextPath it's always outputted in a new line, while, ideally it should be outputted in the same line the path was rendered. 

---

loudenvier:
Added the NoLineBreak flag to TextPath to make it possible to conditionally prevent a new line break from being rendered after the path.

Created Unit tests and added extension method to set the NoLineBreak flag.

Resolves #1177

### DIFF
--- a/src/Spectre.Console/Extensions/TextPathExtensions.cs
+++ b/src/Spectre.Console/Extensions/TextPathExtensions.cs
@@ -116,4 +116,20 @@ public static class TextPathExtensions
     {
         return LeafStyle(obj, new Style(foreground: color));
     }
+
+    /// <summary>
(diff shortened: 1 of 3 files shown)


loudenvier:

PR #1195, opened 2023-03-11
Feature - Add failed task

BatanGaming:
- Add fail state for ProgressTask
- Add failed rendering for: ProgressBar, PercentageColumn, SpinnerColumn, DownloadedColumn

Usage example:
```C#
using Spectre.Console;

await AnsiConsole.Progress()
    .Columns(
        new TaskDescriptionColumn(),
        new ProgressBarColumn
        {
            FailedStyle = new Style(Color.Red),
        },
        new PercentageColumn
        {
            FailedStype = new Style(Color.Red),
        },
        new SpinnerColumn
        {
            CompletedText = "Completed", 
            FailedText = "Failed",
            FailedStyle = new Style(Color.Red),
            CompletedStyle = new Style(Color.Green),
        },
        new DownloadedColumn()
    )
    .AutoClear(false)
    .StartAsync(async context =>
    {
        var goodTask = Task.Run(async () =>
        {
            var task = context.AddTask("Good task");
            var x = 0;
            while (x != 100)
            {
                task.Increment(1);
                x++;
                await Task.Delay(50);
            }
        });

        var badTask = Task.Run(async () =>
        {
            var task = context.AddTask("Bad task");
            try
            {
                var x = 0;
                while (x != 100)
                {
                    task.Increment(1);
                    x++;
                    await Task.Delay(100);
                    if (x == 30)
                    {
                        throw new Exception();
                    }
                }
            }
            catch (Exception ex)
            {
                task.StopTaskWithFailure();
            }
        });
        await Task.WhenAll(goodTask, badTask);
    });
```

Screenshot 2023-03-11 at 10 57 33

### DIFF
--- a/src/Spectre.Console/Live/Progress/Columns/PercentageColumn.cs
+++ b/src/Spectre.Console/Live/Progress/Columns/PercentageColumn.cs
@@ -15,11 +15,25 @@ public sealed class PercentageColumn : ProgressColumn
     /// </summary>
     public Style CompletedStyle { get; set; } = new Style(foreground: Color.Green);
 
+    /// <summary>
+    /// Gets or sets style for a failed task.
+    /// </summary>
+    public Style FailedStype { get; set; } = new Style(foreground: Color.Red);
+
     /// <inheritdoc/>
     public override IRenderable Render(RenderOptions options, ProgressTask task, TimeSpan deltaTime)
     {
         var percentage = (int)task.Percentage;
-        var style = percentage == 100 ? CompletedStyle : Style ?? Style.Plain;
+        var style = Style;
+        if (task.IsFailed)
+        {
+            style = FailedStype;
+        }
+        else if (percentage == 100)
+        {
+            style = CompletedStyle;
+        }
+
         return new Text($"{percentage}%", style).RightJustified();
     }
 
--- a/src/Spectre.Console/Live/Progress/ProgressTask.cs
+++ b/src/Spectre.Console/Live/Progress/ProgressTask.cs
@@ -96,6 +96,11 @@ public sealed class ProgressTask : IProgress<double>
     public bool IsIndeterminate { get; set; }
 
     /// <summary>
+    /// Gets a value indicating whether the task execution failed.
+    /// </summary>
+    public bool IsFailed { get; private set; }
+
+    /// <summary>
     /// Initializes a new instance of the <see cref="ProgressTask"/> class.
     /// </summary>
     /// <param name="id">The task ID.</param>
@@ -149,6 +154,22 @@ public sealed class ProgressTask : IProgress<double>
             StartTime ??= now;
 
             StopTime = now;
+        }
+    }
+
+    /// <summary>
+    /// Stops and marks the task as failed.
+    /// </summary>
+    public void StopTaskWithFailure()
+    {
+        lock (_lock)
+        {
+            var now = DateTime.Now;
+            StartTime ??= now;
+
+            StopTime = now;
+
+            IsFailed = true;
         }
     }
 
--- a/src/Spectre.Console/Widgets/ProgressBar.cs
+++ b/src/Spectre.Console/Widgets/ProgressBar.cs
@@ -5,6 +5,8 @@ internal sealed class ProgressBar : Renderable, IHasCulture
     private const int PULSESIZE = 20;
     private const int PULSESPEED = 15;
 
+    public bool IsFailed { get; set; }
+
     public double Value { get; set; }
     public double MaxValue { get; set; } = 100;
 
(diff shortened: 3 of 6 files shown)


BatanGaming:

src/Spectre.Console/Live/Progress/Columns/PercentageColumn.cs
```diff
@@ -15,11 +15,25 @@ public sealed class PercentageColumn : ProgressColumn
     /// </summary>
     public Style CompletedStyle { get; set; } = new Style(foreground: Color.Green);
 
+    /// <summary>
+    /// Gets or sets style for a failed task.
+    /// </summary>
+    public Style FailedStype { get; set; } = new Style(foreground: Color.Red);
```
thomhurst, line 21:
```suggestion
    public Style FailedStyle { get; set; } = new Style(foreground: Color.Red);
```

BatanGaming: Thanks

src/Spectre.Console/Live/Progress/ProgressTask.cs
```diff
@@ -95,6 +95,11 @@ public double Value
     /// </summary>
     public bool IsIndeterminate { get; set; }
 
+    /// <summary>
+    /// Gets a value indicating whether the task execution failed.
+    /// </summary>
+    public bool IsFailed { get; private set; }
```
redoz, line 101: I would rather see a public setter on this, the current setup with `StopTaskWithFailure` implies the task always stops as soon as it's failed which I don't think is a reasonable assumption to make.

src/Spectre.Console/Widgets/ProgressBar.cs
```diff
@@ -46,7 +49,16 @@ protected override IEnumerable<Segment> Render(RenderOptions options, int maxWid
         }
 
         var bar = !options.Unicode ? AsciiBar : UnicodeBar;
-        var style = isCompleted ? FinishedStyle : CompletedStyle;
+        var style = CompletedStyle;
+        if (isCompleted)
```
redoz, line 53: I'd like to see support for "Completed with errors", something can have run to completion while still encountering errors.

src/Spectre.Console/Live/Progress/ProgressTask.cs
BatanGaming: I don't think so. You named it right `Completed with errors` in the comment below and I think it's not related to `Failed` status. I think what you need is methods like `AddError`, `GetErrors` so then you can get occurred errors. 
I wouldn't mix `Failed` and `Completed` statuses in any way because from my point of view `Failed` task cannot continue working. If some error occurred and task can continue working then I wouldn't call it `Failed`.

redoz: But that feels like adding business-logic into a visual component whose goal it is to represents a process. I just want to visualize "The process encountered an error" and whether that stopped the process, or the process is still going is not related to the visual "error state".

BatanGaming: For me it's just unintuitive to see failed job still going.

@patriksvensson Any chance to review this feature and join the discussion?

redoz: Imagine a CI/CD pipeline, you might have configured it so that even if a test fails, it still publishes the build artifacts. So then the process is in an error state, while still continuing.

In my case the progress bar represents a deployment where a system is deploying 10 different resources, one might fail, but that doesn't stop the deployment from deploying the other nine. (You could argue this is odd behavior, but the behavior is outside of my control, I simply schedule the deployment and monitor its progress).

BatanGaming: Your general example really looks odd for me. But for your particular case, I have something similar actually.
Screenshot 2023-07-26 at 15 55 27
So What I did - I created multiple `ProgressTask` that show the state of each individual task. Since we don't have any option to indicate that error occurred, I made custom widget that shows text state.
After the whole process is over it displays all errors that happened during it.

Back to the topic, my main idea is that 
`Failed` status != `Something bad happened during the process but process still goes on`. 
I agree, it'd be good to have something that indicates that something happened but I think it should be different status, not `Failed`. For your specific case, I'd suggest to separate deployments into independent tasks.

redoz: This is a UI widget why enforce this limitation? If you want to, you can set Failed = true, and then stop the task, and you have what you want, a failed task that is no longer running. 
In my case, I'd like to just say Failed = true and let it run until the underlying task is actually completed indicating that while this is still in progress, it encountered an issue. What's so wrong with having the option to do that?

Issue #1209, opened 2023-04-02
Allow ConfirmationPrompt Styling


wbaldoumas:
**Is your feature request related to a problem? Please describe.**

Right now, `ConfirmationPrompt` styling can't be customized in the same way that `TextPrompt` allows for, which makes for an inconsistent experience when you need to use both and have customized the `TextPrompt` styling.

**Describe the solution you'd like**

I'd like to introduce the same styling capabilities to `ConfirmationPrompt` that `TextPrompt` currently has.

**Describe alternatives you've considered**

For now, I've been using a custom [`ConfirmationPromptWithStyling`](https://github.com/wbaldoumas/graphql-to-karate/blob/main/src/GraphQLToKarate.CommandLine/Prompts/ConfirmationPromptWithStyling.cs) in my own library that needs it, but it would be great if this came with this library so I can delete this custom implementation. 😄 

**Additional context**
N/A

---
Please upvote :+1: this issue if you are interested in it.

PR #1210, opened 2023-04-02
Allow Confirmation Prompt Styling

wbaldoumas:
Allows styling of `ConfirmationPrompt` in the same way that `TextPrompt` can be styled for choices and default values. 

If no style is provided, fallback to the same `"blue"` and `"green"` that `TextPrompt` uses is done.

Addresses https://github.com/spectreconsole/spectre.console/issues/1209.

### DIFF
--- a/src/Spectre.Console/Prompts/ConfirmationPrompt.cs
+++ b/src/Spectre.Console/Prompts/ConfirmationPrompt.cs
@@ -40,6 +40,16 @@ public sealed class ConfirmationPrompt : IPrompt<bool>
     public bool ShowDefaultValue { get; set; } = true;
 
     /// <summary>
+    /// Gets or sets the style in which the default value is displayed.
+    /// </summary>
+    public Style? DefaultValueStyle { get; set; }
+
+    /// <summary>
+    /// Gets or sets the style in which the list of choices is displayed.
+    /// </summary>
+    public Style? ChoicesStyle { get; set; }
+
+    /// <summary>
     /// Gets or sets the string comparer to use when comparing user input
     /// against Yes/No choices.
     /// </summary>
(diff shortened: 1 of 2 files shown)


wbaldoumas:

wbaldoumas:
@patriksvensson curious about your thoughts on this one and if there are any issues with the proposed changes here. 😄

src/Spectre.Console/Prompts/ConfirmationPrompt.cs
```diff
@@ -39,6 +39,16 @@ public sealed class ConfirmationPrompt : IPrompt<bool>
     /// </summary>
     public bool ShowDefaultValue { get; set; } = true;
 
+    /// <summary>
+    /// Gets or sets the style in which the default value is displayed.
```
0xced, line 43:
```suggestion
    /// Gets or sets the style in which the default value is displayed. Defaults to green when <see langword="null"/>.
```

src/Spectre.Console/Prompts/ConfirmationPrompt.cs
```diff
@@ -39,6 +39,16 @@ public sealed class ConfirmationPrompt : IPrompt<bool>
...
+    /// </summary>
+    public Style? DefaultValueStyle { get; set; }
+
+    /// <summary>
+    /// Gets or sets the style in which the list of choices is displayed.
```
0xced, line 48:
```suggestion
    /// Gets or sets the style in which the list of choices is displayed. Defaults to blue when <see langword="null"/>.
```

0xced (commented):
Seems like a good addition. I have proposed some changes that will be applicable once #1359 is merged.

wbaldoumas:
@0xced applied your suggestions. Thanks for the review!

0xced (approved):
Thanks for your contribution!

PR #1212, opened 2023-04-04
Added verification overloading to AnsiConsole.Prompt

aijkl:
There is a case where we would like to add validation to AnsiConsole.Ask, we have added an overload, please consider merging

```
        var age = AnsiConsole.Ask<int>("How old are you?", i =>
        {
            return i > 200 ? ValidationResult.Error("Too big.") : ValidationResult.Success();
        });
```
image


aijkl:

aijkl:
@patriksvensson I apologize for the inconvenience, but I would appreciate a review.

PR #1241, opened 2023-06-12
Add fix to avoid exception on Rows with no children

jeppevammenkristensen:
I ran into an issue where intialized and added Rows with no child renderables, you got an exception when tried to render it. 

`System.InvalidOperationException: Sequence contains no elements
   at System.Linq.ThrowHelper.ThrowNoElementsException()
   at System.Linq.Enumerable.MinInteger[TSource,TResult](IEnumerable`1 source, Func`2 selector)
   at System.Linq.Enumerable.Min[TSource](IEnumerable`1 source, Func`2 selector)
   at Spectre.Console.Rows.Measure(RenderOptions options, Int32 maxWidth) in /_/src/Spectre.Console/Widgets/Rows.cs:line 41
   at Spectre.Console.Rendering.Renderable.Spectre.Console.Rendering.IRenderable.Measure(RenderOptions options, Int32 maxWidth) in /_/src/Spectre.Console/Rendering/Renderable.cs:line 12` 

The error occured because we call the Min extension method, that will throw an exception if the IEnumerable input is empty. 

In this small fix we check to ensure if measurements is empty, and in that case return a mesurement with min and max set to 0.


jeppevammenkristensen:
> @microsoft-github-policy-service agree

jeppevammenkristensen:

patriksvensson (approved):
LGTM! 👍

PR #1252, opened 2023-07-05
Help writer improvements

FrankRay78:
**Summary of changes**

Version option will show in help even with a default command

Reserve `-v` and `--version` as special spectre.console command line arguments (nb. breaking change for spectre.console users who have a default command with a settings class that uses either of these switches)

Help writer correctly determines if trailing commands exist and whether to display them as optional or mandatory in the usage statement

Ability to control the number of indirect commands to display in the help text when the command itself doesn't have any examples of its own. Defaults to 5 (for backward compatibility) but can be set to any integer or zero to disable completely.

Significant increase in unit test coverage for the help writer

Minor grammatical improvements to website documentation


FrankRay78:
**DON'T MERGE THIS PR**
I propose to merge the changes in this PR through it's ancestor PR https://github.com/spectreconsole/spectre.console/pull/1259  (once merged, I will delete this PR)

patriksvensson:
@FrankRay78 I've converted this PR to a draft so it won't be accidentally merged

Issue #1234, opened 2023-05-25
Possibility to highlight calendar events in different colors


SeppPenner:
**Is your feature request related to a problem? Please describe.**
We use the Spectre.Console calendar and wanted to highlight certain days in different colors. E.g. think of a work week (Monday to Friday is e.g. green, Saturday and Sunday would be e.g. yellow).

**Describe the solution you'd like**
The possibility to use the `HighlightStyle()` function on a per-day basis with the calendar. 

**Describe alternatives you've considered**
None.

**Additional context**
Nothing special here.

---
Please upvote :+1: this issue if you are interested in it.

PR #1246, opened 2023-06-22
Add custom style for each calendar event

davide-pi:
# Description

Implemented the custom highlight style for single calendar event.
Also requested with issue #1234.

# Example

As shown below the calendar now support the default highlight style `2023-06-22` but also support different style for different events like `2023-06-24` and `2023-06-25` in yellow foreground and `2023-06-27` with red foreground

image

# To Test

To test this implementation is sufficient to put it into the `Main` method of the `Program.cs` in the `Spectre.Console.Analyzer.Sandbox` .

```csharp
        var today = DateTime.Today;

        var calendar = new Calendar(today.Year, today.Month);

        //Just keep an eye on the current day
        calendar.AddCalendarEvent(today);

        //Interesting events
        calendar.AddCalendarEvent(today.AddDays(2), Color.Yellow);
        calendar.AddCalendarEvent(today.AddDays(3), Color.Yellow);

        //Inderogable important event
        calendar.AddCalendarEvent(today.AddDays(5), Color.Red);

        AnsiConsole.Write(calendar);
```

---
Please upvote :+1: this pull request if you are interested in it.

PR #1254, opened 2023-07-10
Added option to set the calendar highlight style on a daily basis.

SeppPenner:
Fixes https://github.com/spectreconsole/spectre.console/issues/1234.

However, this would be a breaking change as the `calendar.HighlightStyle()` method is completely removed (https://spectreconsole.net/widgets/calendar). --> Maybe the docs need to be adjusted afterwards as well.


SeppPenner:

SeppPenner:
Closing it in favor of https://github.com/spectreconsole/spectre.console/pull/1246.

PR #1255, opened 2023-07-12
Remove unnecessary [NotNull] attributes

0xced:
When subclassing `Command<TSettings>` which has the [NotNull] attributes, Rider/ReSharper gives this warning:
> Nullability of type of parameter 'context' in method does not match overridden member `int Spectre.Console.Cli.Command<TSettings>.Execute(CommandContext, TSettings)` (possibly because of nullability attributes)

image

When subclassing `Command<TSettings>` which does not have the [NotNull] attributes, Rider/ReSharper gives this warning:
> The nullability attribute has no effect and can be safely removed

image

The solution is simply to remove the [NotNull] attributes.

Since `<Nullable>enable</Nullable>` is set in the project, they are actually not necessary. By the way, the non-generic `Command` class does not have the [NotNull] attributes.


patriksvensson:
Thanks @0xced! Merged! 👍

PR #1259, opened 2023-07-19
Allow custom help providers

FrankRay78:
**Work done**

- Implement IHelpProvider
- Port existing HelpWriter to use this new interface
- Remove coupling to internal sealed classes
- Set existing help writer as the default help provider
- Implement a custom help provider to demonstrate functionality and pass unit tests

**Notes**

1. This PR was branched from https://github.com/spectreconsole/spectre.console/pull/1252, which seemed like a reasonable starting place.


rcdailey:
I'll look through this, but the change set is really large (I see 54 files)! If there's any possibility to split the PR to make things easier to review, that would be appreciated, but I will make do with what you have for now.

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -8,85 +8,97 @@ public CommandExecutor(ITypeRegistrar registrar)
     {
         _registrar = registrar ?? throw new ArgumentNullException(nameof(registrar));
         _registrar.Register(typeof(DefaultPairDeconstructor), typeof(DefaultPairDeconstructor));
-    }
-
+    }
+
+#pragma warning disable SA1507 // Code should not contain multiple blank lines in a row
```
rcdailey, line 13: Why are the multiple blank lines important?

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -8,85 +8,97 @@ public CommandExecutor(ITypeRegistrar registrar)
...
+
+        args ??= new List<string>();
+
+
+        _registrar.RegisterInstance(typeof(IConfiguration), configuration);
+        _registrar.RegisterLazy(typeof(IAnsiConsole), () => configuration.Settings.Console.GetConsole());
+
+
+        // Register the default help provider
+        var defaultHelpProvider = new Help.DefaultHelpProvider(configuration.Settings.ShowOptionDefaultValues, configuration.Settings.MaximumIndirectExamples, configuration.Settings.TrimTrailingPeriod);
```
rcdailey, line 29: Why not pass in `configuration.Settings` and let `DefaultHelpProvider` decide which settings it wants? I could see an argument for making these an implementation detail.

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -8,85 +8,97 @@ public CommandExecutor(ITypeRegistrar registrar)
...
+        _registrar.RegisterInstance(typeof(Help.IHelpProvider), defaultHelpProvider);
```
rcdailey, line 30: Bear with me on this one; I don't fully understand the architecture of Spectre.Console yet. I did check out your PR locally so I can poke around.

This `Execute()` method gets invoked when `Run()` is called. My first question is: Can we avoid doing registrations at this point? IMO, registrations should happen before application code execution. In this case, moving it to the constructor would give the effect I'm looking for. However, I assume it's here for a reason that I don't fully understand yet.

Another question I have is: What happens if a user registers their own `IHelpProvider`? Shouldn't we skip the construction of `DefaultHelpProvider` as well as its registration?

rcdailey (commented):
Whew, I got through some of this but there is a lot to go. I need to get back to work but I'll share the comments I have so far. Hopefully I didn't overwhelm you too much. So far, I don't see any issues I feel really strongly about, so please consider all of my comments as unimportant.

Thanks for your effort and giving me the opportunity to review and learn more about the code base.

FrankRay78:
Thanks for the review @rcdailey, much appreciated.  There are a number of things I want to address here, however I'm on holiday for a few weeks shortly and so will pick them up when back mid-Aug.  Some of the review issues are hangovers from the existing code base, rather than the introduction of an injectible HelpWriter, however now is probably a good time to consider improving matters.

src/Spectre.Console.Cli/Help/ICommandInfoExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        var current = commandInfo;
+        while (current != null)
+        {
+            result.Push(current);
+            current = current.Parent;
+        }
+
+        return result.ToList();
+    }
```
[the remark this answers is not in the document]
elgonzo: @rcdailey 

Your suggestion is not exhibiting equivalent behavior.

Pay attention to how the code in the PR uses a stack. The ToList() method uses the stack's enumerator, which enumerates the stack from bottom to up (from last pushed element to first pushed element). Consider how the stack is populated, and the code in the PR delivers a list in which the commands are ordered beginning with the root command followed by the child sub commands that will lead to the sub command provided to the function.

Your code however will result in exactly the opposite ordering, an enumerable that begins with the provided sub command and then with it's successive parent commands until it reaches root.

src/Spectre.Console.Cli/Help/DefaultHelpProvider.cs
```diff
@@ -46,49 +57,62 @@ public HelpOption(string? @short, string? @long, string? @value, bool? valueIsOp
...
-            parameters.Add(new HelpOption("h", "help", null, null, "Prints help information", null));
-
-            // At the root and no default command?
-            if (command == null && model?.DefaultCommand == null)
-            {
-                parameters.Add(new HelpOption("v", "version", null, null, "Prints version information", null));
+            parameters.Add(new HelpOption("h", "help", null, null, "Prints help information", null));
+
+            // At the root?
+            if ((command == null || command?.Parent == null) && !(command?.IsBranch ?? false))
```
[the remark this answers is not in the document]
FrankRay78: I have given this some decent throught, and felt the best way to address this issue was to actually improve the documentation at line 65, the most recent push includes:

```csharp
            // Version information applies to the entire application
            // Include the "-v" option in the help when at the root of the command line application
            // Don't allow the "-v" option if users have specified one or more sub-commands
            if ((command == null || command?.Parent == null) && !(command?.IsBranch ?? false))
            {
```
I looked at the extension method suggestion and didn't feel that it fitted very well, namely:
1. `command == null` check would remain as is
2. `command?.Parent == null` is what the `IsRoot` property would be implementing
3. `!(command?.IsBranch ?? false)` would remain as is, on the if check (rather than within the `IsRoot` property) as whether the command is a branch or not, does not relate to whether we are at the root or not

src/Spectre.Console.Cli/Help/ICommandInfoExtensions.cs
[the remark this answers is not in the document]
FrankRay78: Good suggestion @rcdailey (and setting aside @elgonzo's comment above), but my preference would be to leave it ungeneralised, as this is how it's currently implemented in the main branch prior to this PR. I've just relocated the method. 

I also feel it's more readable 'as is', although that is only a personal opinion and would not be sufficient by itself to avoid improving the codebase.

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
FrankRay78: The `CommandExecutor.Execute(...)` method was already large and a potential refactoring candidate. 

As I wasn't making significant changes to the method in this PR, nor did I want to refactor it extensively, I did however introduce some double lines (see arrows) to start to better group together distinct pieces of functionality.

rcdailey:
Thanks for the replies. Given how hard you've worked on this and the fact that I haven't really provided any feedback that is terribly important, I don't want to hold you up any more on this PR. There's also a lot to go through and I've had difficulty finding time to resume where I left off.

All that to say: I think what you have is going to be just fine, so if you happen to merge this without any further comments from me, I think that's just fine. Thanks again for all of your hard work.

rcdailey (approved):
Funny enough, as soon as I leave a comment saying I had no time to review the rest... I ended up making time to do it anyway! Looks like for the most part all I had left to look at was some test code / data files, which was easy. I'll go ahead and give this my approval, for whatever that's worth. Great job!!

EDIT: I'll just add: I think the *only* thing I have somewhat strong feelings about is my comment from before about mixing DI registrations in the post-initialization / runtime code. I'm sure there are architectural reasons for it, otherwise I imagine you'd have done it differently. We can discuss it in the original comment. I just wanted to put an asterisk next to my approval by reminding you about that one thing.

FrankRay78:
Hi @rcdailey, I've really appreciated your review comments so far.  What's really impressive is actually your last comment about DI, particularly since you aren't 100% familiar with the spectre.console codebase.  I've left that review comment until last, picking off the easier ones first, knowing that (in good faith) I should really take the time to consider addressing the DI registration issue (if I can). 

You've seen the slightly "bodged" use of how  the following line still gets called, `_registrar.RegisterInstance(typeof(Help.IHelpProvider), defaultHelpProvider);` irrespective of whether a custom `IHelpProvider` was already registered firstly. When coding this, I discovered the existing DI container allows multiple registrations for the same interface 😱, always resolving to the first one. 

I've been thinking about having it throw an exception, or perhaps replacing the existing. It's this I want to spend a bit more time on before having this PR merged.

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
FrankRay78: I think I agree with this actually, I'm going to code it now and see what it looks like...

rcdailey:
> I discovered the existing DI container allows multiple registrations for the same interface 😱, always resolving to the first one

I am not sure which DI container you're referring to, but it probably doesn't matter much since Spectre.Console is agnostic to the DI implementation. For example, I use Autofac, and it actually sets the *last* registration as the default, not the first. There's a [mechanism to override that](https://docs.autofac.org/en/latest/register/registration.html#default-registrations), though.

Would it make sense to use a configuration property for this instead of letting a user register an implementation "behind your back"?

```cs
app.Configure(config =>
{
    config.SetHelpProvider<MyCustomHelp>();
    config.ValidateExamples();
});
```

This would allow you to conditionally register this type using `ITypeRegistrar` (either a default, as you currently do, or the user-provided one. And you'd be able to do this at "configure" time instead of in the `Run()` execution path.

I'm sure there's a better way, and certainly more than one way. This is food for thought at least.

FrankRay78:
> Would it make sense to use a configuration property for this instead

That's a good idea for further consideration, given we are not quite at the stage of wholesale allowing any interface to be overridden and injected. And this PR is focused on specifically allowing custom help providers.

The current DI is custom (before my time), basically to avoid / reduce dependencies on third-party assemblies, see:

image

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
FrankRay78: I've now pushed changes to make DefaultHelpProvider constructor accept an ICommandAppSettings

src/Spectre.Console.Cli/Help/ICommandArgument.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Gets the value of the argument.
+    /// </summary>
+    string Value { get; }
+
+    /// <summary>
+    /// Gets or sets the position of the argument.
+    /// </summary>
+    int Position { get; set; }
```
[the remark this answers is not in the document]
FrankRay78: Good query.  The setter is used in `CommandModelBuilder`:

image

rcdailey:
> this PR is focused on specifically allowing custom help providers

I must have a fundamental misunderstanding, then. I thought the way users added custom help providers was by implementing `IHelpProvider`. What am I missing?

src/Spectre.Console.Cli/Help/ICommandArgument.cs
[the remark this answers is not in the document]
rcdailey: The setter of `CommandArgument.Position` (the concrete type) is used there, but not `ICommandArgument.Position` (the interface). I checked out your branch and removed the setter for `ICommandArgument.Position` and I got no compiler errors. 

Is there some other use case where `ICommandArgument.Position` needs to be assignable?

FrankRay78:
> I must have a fundamental misunderstanding, then. I thought the way users added custom help providers was by implementing IHelpProvider. What am I missing?

Sorry, I meant the various internally marked interfaces that spectre.console currently uses.  This PR, which is entirely focused on allowing users to extend IHelpProvider, is a beachhead for considering opening up _some_ of the other interfaces for DI, in the fullness of time.  

> IMO, registrations should happen before application code execution

this becomes important to get right, which I will look at shortly.  If i can refactor this in the current PR, I will. Otherwise I'll get the PR merged and the extensible help into main, then come back to do the remaining DI work on a separate PR.

Hope that clears up my thinking.  Thanks for chatting this through.

rcdailey:
> Sorry, I meant the various internally marked interfaces that spectre.console currently uses. This PR, which is entirely focused on allowing users to extend IHelpProvider,

Ok then we're on the same page, I think. What confuses me is your response to my idea about passing in `IHelpProvider` via the configuration object. Your response to that was this:

> That's a good idea for further consideration, given we are not quite at the stage of wholesale allowing any interface to be overridden and injected

Maybe you thought I was talking about a different interface (probably an internal one)? I was only talking about `IHelpProvider`. I just want to make sure we're not talking past each other. I think we're both on the same subject, though.

Thinking about the default help provider a little more. Currently, I think you have it working like this (in `CommandExecutor.Execute()`):

1. Create a `DefaultHelpProvider` instance
2. Register that instance with `ITypeRegistrar`
3. Attempt to resolve `IHelpProvider`
4. If the resolution returns `null`, fall back to the `DefaultHelpProvider` instance created in step 

Is step 2 necessary? In other words, is there a scenario where if a user does *not* register a custom implementation of `IHelpProvider` that something else outside of the Spectre.Console code base would want to resolve it?

If we can get rid of step 2, then I think that fixes the issue I brought up earlier about a user being potentially unable to override the default help provider implementation due to multiple registrations of `IHelpProvider`. This would also allow you to avoid constructing a `DefaultHelpProvider` until you need one:

```cs
// Get the registered help provider, falling back to the default provider
// registered above if no custom implementations have been registered.
var helpProvider = resolver.Resolve(typeof(Help.IHelpProvider)) as Help.IHelpProvider 
    ?? new DefaultHelpProvider(configuration.Settings);
```

FrankRay78:
We are on the same page @rcdailey, 

However, the bad memories are coming back now...  this line `resolver.Resolve(typeof(Help.IHelpProvider)) as Help.IHelpProvider ` actually throws an exception if an IHelpProvider hasn't been registered. 

This is because the `TypeResolverAdapter` class falls back to using the .Net activator, namely `return Activator.CreateInstance(type);`, which throws because it can't instantiate an interface.

What I think I need to do is add a `.Contains(type)` method to `DefaultTypeRegistrar`, so the `CommandExecutor` can explicitly check if an `IHelpProvider` has been registered, before deciding whether to call `resolver.Resolve(typeof(Help.IHelpProvider))`, or alternatively just instantiate the `DefaultHelpProvider`.

I'm going to see what this looks like right now...

src/Spectre.Console.Cli/Help/ICommandArgument.cs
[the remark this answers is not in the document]
FrankRay78: Good spot, I missed this.  I'll push an update now.

rcdailey:
Sounds good. I assume by adding it to `DefaultTypeRegistrar`, you'll also be adding it to the `ITypeRegistrar` itnerface? I assume you'll want the ability to verify a registration already exists for custom implementations that users provided (as I mentioned before, I implement it for Autofac).

If that doesn't work out due to the user-observable changes, you could do the below. I don't like it personally, because it sort of toes the line on using exceptions for flow control. But it's better than nothing...

```cs
IHelpProvider helpProvider = null;
try 
{
  helpProvider = resolver.Resolve(typeof(Help.IHelpProvider)) as Help.IHelpProvider;
}
finally
{
  helpProvider ??= new DefaultHelpProvider(configuration.Settings);
}
```

Whatever you decide, I'm all for it. Just sharing an idea. Thanks again for the hard work!!

FrankRay78:
I've looked at this and have a heap more uncommitted changes to clean up/refactor/improve the DI. 

However, given the size of this PR already, I'm proposing that we leave this one here and seek to get it merged into main. The review has been high-quality and I'm happy with everything checked in currently.

The current way the default help provider is created, then registered, then referenced lower down (which isn't great) is consistent with other patterns in the same method, see red circled below: 
image

I think I would prefer to open a new PR to address the DI-specific refactoring, including looking at splitting the `CommandExecutor.Execute()` method (see yellow arrow and comment) into something like `.FinaliseRegistrations` and then `.Execute`, rather than complicate this PR further. 

Are you in agreement with this @rcdailey ?

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
FrankRay78: I propose completing this in a separate, follow on PR, as explained in the comment here:  https://github.com/spectreconsole/spectre.console/pull/1259#issuecomment-1666942715

FrankRay78:
**Update:**

I've added the ability to set the help provider via the configurator (thanks for the suggestion @rcdailey), ie:

image

and

image

(nb. the instance needs to be passed like this for any providers that require constructor arguments, at least until the DI supports constructor injection. Custom providers could then potentially be instantiated with any other class that has been registered, `ICommandAppSettings` being a good example)

I have also made the internals of the `DefaultHelpProvider` as `public virtual`, and normalised the method parameters for consistency and to provide easy access to them in each of the methods ie.:

image

Which significantly helps users wishing to add their own minor tweaks to the current help writer functionality, ie. adding a custom header to the help information simply becomes:

image

rcdailey:
Sorry for the late response!

Frank, you have done an amazing job in this PR. Regarding your previous reply, I think addressing the DI rework in a separate PR is a great idea. Let's take this one as a win!

I also really love the changes you've made for both specifying a custom help provider and allowing small tweaks to default behavior!

For what it's worth, I will give this one my approval :)

rcdailey (approved):

FrankRay78:
Code complete & peer-reviewed @patriksvensson, including:

- New project added to the examples solution
- New documentation page added to the website

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -113,7 +125,8 @@ public async Task<int> Execute(IConfiguration configuration, IEnumerable<string>
 
         return parsedResult;
     }
-
+#pragma warning restore CS8603 // Possible null reference return.
```
patriksvensson, line 128: Remove pragma

src/Spectre.Console.Cli/Internal/Modelling/CommandArgument.cs
```diff
@@ -1,6 +1,6 @@
 namespace Spectre.Console.Cli;
 
-internal sealed class CommandArgument : CommandParameter
+internal sealed class CommandArgument : CommandParameter, Spectre.Console.Cli.Help.ICommandArgument
```
patriksvensson, line 3: Qualify the namespace

src/Spectre.Console.Cli/Internal/Modelling/CommandInfo.cs
```diff
@@ -1,6 +1,7 @@
 namespace Spectre.Console.Cli;
-
-internal sealed class CommandInfo : ICommandContainer
+
+[SuppressMessage("StyleCop.CSharp.NamingRules", "SA1124:DoNotUseRegions", Justification = "Hiding Help.ICommandInfo explicit interface implementation in a region improves readability.")]
+internal sealed class CommandInfo : ICommandContainer, Spectre.Console.Cli.Help.ICommandInfo
```
patriksvensson, line 4: Qualify the namespace

src/Spectre.Console.Cli/Internal/Modelling/CommandInfo.cs
```diff
@@ -20,8 +21,17 @@ internal sealed class CommandInfo : ICommandContainer
 
     // only branches can have a default command
     public CommandInfo? DefaultCommand => IsBranch ? Children.FirstOrDefault(c => c.IsDefaultCommand) : null;
-    public bool IsHidden { get; }
-
+    public bool IsHidden { get; }
+
+    #region Help.ICommandInfo
+
+    IList<Help.ICommandInfo> Help.ICommandContainer.Commands => Children.Cast<Help.ICommandInfo>().ToList();
```
patriksvensson, line 28:
```suggestion
    IList<ICommandInfo> ICommandContainer.Commands => Children.Cast<ICommandInfo>().ToList();
```

src/Spectre.Console.Cli/Internal/Modelling/CommandInfo.cs
```diff
@@ -20,8 +21,17 @@ internal sealed class CommandInfo : ICommandContainer
...
     // only branches can have a default command
     public CommandInfo? DefaultCommand => IsBranch ? Children.FirstOrDefault(c => c.IsDefaultCommand) : null;
-    public bool IsHidden { get; }
-
+    public bool IsHidden { get; }
+
+    #region Help.ICommandInfo
+
+    IList<Help.ICommandInfo> Help.ICommandContainer.Commands => Children.Cast<Help.ICommandInfo>().ToList();
+    Help.ICommandInfo? Help.ICommandContainer.DefaultCommand => DefaultCommand;
```
patriksvensson, line 29:
```suggestion
    ICommandInfo? ICommandContainer.DefaultCommand => DefaultCommand;
```

src/Spectre.Console.Cli/Internal/Modelling/CommandInfo.cs
```diff
@@ -20,8 +21,17 @@ internal sealed class CommandInfo : ICommandContainer
...
+    IList<Help.ICommandParameter> Help.ICommandInfo.Parameters => Parameters.Cast<Help.ICommandParameter>().ToList();
```
patriksvensson, line 30:
```suggestion
    IList<ICommandParameter> ICommandInfo.Parameters => Parameters.Cast<ICommandParameter>().ToList();
```

src/Spectre.Console.Cli/Internal/Modelling/CommandInfo.cs
```diff
@@ -20,8 +21,17 @@ internal sealed class CommandInfo : ICommandContainer
...
+    Help.ICommandInfo? Help.ICommandInfo.Parent => Parent;
```
patriksvensson, line 31:
```suggestion
    ICommandInfo? ICommandInfo.Parent => Parent;

```

src/Spectre.Console.Cli/Internal/Modelling/CommandModel.cs
```diff
@@ -1,14 +1,22 @@
 namespace Spectre.Console.Cli;
-
-internal sealed class CommandModel : ICommandContainer
+
+[SuppressMessage("StyleCop.CSharp.NamingRules", "SA1124:DoNotUseRegions", Justification = "Hiding Help.ICommandModel explicit interface implementation in a region improves readability.")]
```
patriksvensson, line 3: Remove suppression

src/Spectre.Console.Cli/Internal/Modelling/CommandModel.cs
```diff
@@ -1,14 +1,22 @@
...
-    public CommandInfo? DefaultCommand => Commands.FirstOrDefault(c => c.IsDefaultCommand);
+    public CommandInfo? DefaultCommand => Commands.FirstOrDefault(c => c.IsDefaultCommand);
+
+    #region Help.ICommandModel
+
+    string Help.ICommandModel.ApplicationName => GetApplicationName();
+    IList<Help.ICommandInfo> Help.ICommandContainer.Commands => Commands.Cast<Help.ICommandInfo>().ToList();
+    Help.ICommandInfo? Help.ICommandContainer.DefaultCommand => DefaultCommand;
+
+    #endregion
```
patriksvensson, line 19: No regions

src/Spectre.Console.Cli/Internal/Modelling/CommandModel.cs
```diff
@@ -1,14 +1,22 @@
...
     public IList<CommandInfo> Commands { get; }
     public IList<string[]> Examples { get; }
-    public bool TrimTrailingPeriod { get; }
 
-    public CommandInfo? DefaultCommand => Commands.FirstOrDefault(c => c.IsDefaultCommand);
+    public CommandInfo? DefaultCommand => Commands.FirstOrDefault(c => c.IsDefaultCommand);
+
+    #region Help.ICommandModel
+
+    string Help.ICommandModel.ApplicationName => GetApplicationName();
```
patriksvensson, line 15:
```suggestion
    string ICommandModel.ApplicationName => GetApplicationName();
```

src/Spectre.Console.Cli/Internal/Modelling/CommandModel.cs
```diff
@@ -1,14 +1,22 @@
...
+    IList<Help.ICommandInfo> Help.ICommandContainer.Commands => Commands.Cast<Help.ICommandInfo>().ToList();
```
patriksvensson, line 16:
```suggestion
    IList<Help.ICommandInfo> Help.ICommandContainer.Commands => Commands.Cast<Help.ICommandInfo>().ToList();
```
```suggestion
    IList<ICommandInfo> ICommandContainer.Commands => Commands.Cast<ICommandInfo>().ToList();
```

src/Spectre.Console.Cli/Internal/Modelling/CommandModel.cs
```diff
@@ -1,14 +1,22 @@
...
+    Help.ICommandInfo? Help.ICommandContainer.DefaultCommand => DefaultCommand;
```
patriksvensson, line 17:
```suggestion
    ICommandInfo? ICommandContainer.DefaultCommand => DefaultCommand;
```

src/Spectre.Console.Cli/Internal/Modelling/CommandOption.cs
```diff
@@ -1,6 +1,6 @@
 namespace Spectre.Console.Cli;
 
-internal sealed class CommandOption : CommandParameter
+internal sealed class CommandOption : CommandParameter, Spectre.Console.Cli.Help.ICommandOption
```
patriksvensson, line 3:
```suggestion
internal sealed class CommandOption : CommandParameter, ICommandOption
```

src/Spectre.Console.Cli/Internal/Modelling/CommandParameter.cs
```diff
@@ -1,6 +1,7 @@
 namespace Spectre.Console.Cli;
-
-internal abstract class CommandParameter : ICommandParameterInfo
+
+[SuppressMessage("StyleCop.CSharp.NamingRules", "SA1124:DoNotUseRegions", Justification = "Hiding Help.ICommandParameter explicit interface implementation in a region improves readability.")]
```
patriksvensson, line 3: Remove suppression

src/Spectre.Console.Cli/Internal/Modelling/CommandParameter.cs
```diff
@@ -1,6 +1,7 @@
...
+internal abstract class CommandParameter : ICommandParameterInfo, Spectre.Console.Cli.Help.ICommandParameter
```
patriksvensson, line 4:
```suggestion
internal abstract class CommandParameter : ICommandParameterInfo, ICommandParameter
```

src/Spectre.Console.Cli/Internal/Modelling/CommandParameter.cs
```diff
@@ -17,8 +18,14 @@ internal abstract class CommandParameter : ICommandParameterInfo
...
     public virtual bool WantRawValue => ParameterType.IsPairDeconstructable()
-        && (PairDeconstructor != null || Converter == null);
-
+        && (PairDeconstructor != null || Converter == null);
+
+    #region Help.ICommandParameter
+
+    public bool IsFlag => ParameterKind == ParameterKind.Flag;
+
+    #endregion
```
patriksvensson, line 27: No regions

test/Spectre.Console.Cli.Tests/Data/Help/CustomHelpProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+            new Text("--------------------------------------"), Text.NewLine,
+            new Text("---      CUSTOM HELP PROVIDER      ---"), Text.NewLine,
+            new Text("--------------------------------------"), Text.NewLine,
+            Text.NewLine,
+        };
+    }
+
+    public override IEnumerable<IRenderable> GetFooter(ICommandModel model, ICommandInfo command)
+    {
+        return new[]
```
patriksvensson, line 29:
```suggestion
        return new IRenderable[]
```

examples/Cli/Help/Help.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
patriksvensson, line 7:
```suggestion
    <Nullable>enable</Nullable>
    <ExampleName>Help</ExampleName>
    <ExampleDescription>Some description for help</ExampleDescription>
    <ExampleGroup>Cli</ExampleGroup>
    <ExampleVisible>false</ExampleVisible>
```

CLI examples are not really working for `dotnet example`, so we need to add the metadata and exclude it from appear in the listing.

patriksvensson (changes requested):
An initial review of the code. Everything looks good what I can see, but there are some things I think need to be fixed.

One thing is that we probably shouldn't put help stuff in its own namespace unless we want to hide it by default. If we put it in a namespace, we should add `Spectre.Console.Cli.Help` to `Usings.cs` so we don't have to explicitly qualify the namespace.

src/Spectre.Console.Cli/Help/ICommandModel.cs
```diff
@@ -0,0 +1,12 @@
...
+
+/// <summary>
+/// Represents a command model.
+/// </summary>
+public interface ICommandModel : ICommandContainer
+{
+    /// <summary>
+    /// Gets the name of the application.
+    /// </summary>
+    string ApplicationName { get; }
```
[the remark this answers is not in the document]
FrankRay78: CommandModel always provides a non-null application name when cast as a ICommandModel (as used by the help provider). It's the following logic that ensures this:

```csharp
string ICommandModel.ApplicationName => GetApplicationName(ApplicationName);

// [snip snip]

    /// <summary>
    /// Gets the name of the application.
    /// If the provided <paramref name="applicationName"/> is not null or empty,
    /// it is returned. Otherwise the name of the current application
    /// is determined based on the executable file's name.
    /// </summary>
    private static string GetApplicationName(string? applicationName)
    {
        return
            applicationName ??
            Path.GetFileName(GetApplicationFile()) ?? // null is propagated by GetFileName
            "?";
    }
```

examples/Cli/Help/DefaultCommand.cs
```diff
@@ -0,0 +1,17 @@
+using Spectre.Console;
+using Spectre.Console.Cli;
+
+namespace Help;
+
+public sealed class DefaultCommand : Command
+{
+    public DefaultCommand()
```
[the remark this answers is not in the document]
FrankRay78: Done

examples/Cli/Help/CustomHelpProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+internal class CustomHelpProvider : DefaultHelpProvider
+{
+    public CustomHelpProvider(ICommandAppSettings settings)
+        : base(settings)
+    {
+    }
+
+    public override IEnumerable<IRenderable> GetHeader(ICommandModel model, ICommandInfo? command)
+    {
+        return new[]
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/ConfiguratorExtensions.cs
```diff
@@ -5,7 +5,34 @@ namespace Spectre.Console.Cli;
...
+{
+    /// <summary>
+    /// Sets the help provider for the application.
+    /// </summary>
+    /// <param name="configurator">The configurator.</param>
+    /// <param name="helpProvider">The help provider to use.</param>
+    /// <returns>A configurator that can be used to configure the application further.</returns>
+    public static IConfigurator SetHelpProvider(this IConfigurator configurator, Help.IHelpProvider helpProvider)
+    {
+        configurator.SetHelpProvider(helpProvider);
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/ConfiguratorExtensions.cs
```diff
@@ -5,7 +5,34 @@ namespace Spectre.Console.Cli;
...
+
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/ConfiguratorExtensions.cs
```diff
@@ -5,7 +5,34 @@ namespace Spectre.Console.Cli;
...
+    /// <summary>
+    /// Sets the help provider for the application.
+    /// </summary>
+    /// <param name="configurator">The configurator.</param>
+    /// <typeparam name="T">The type of the help provider to instantiate at runtime and use.</typeparam>
+    /// <returns>A configurator that can be used to configure the application further.</returns>
+    public static IConfigurator SetHelpProvider<T>(this IConfigurator configurator)
+        where T : Help.IHelpProvider
+    {
+        configurator.SetHelpProvider<T>();
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/ConfiguratorExtensions.cs
```diff
@@ -5,7 +5,34 @@ namespace Spectre.Console.Cli;
...
+
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandInfo.cs
```diff
@@ -20,8 +21,17 @@ internal sealed class CommandInfo : ICommandContainer
 
     // only branches can have a default command
     public CommandInfo? DefaultCommand => IsBranch ? Children.FirstOrDefault(c => c.IsDefaultCommand) : null;
-    public bool IsHidden { get; }
-
+    public bool IsHidden { get; }
+
+    #region Help.ICommandInfo
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -8,85 +8,97 @@ public CommandExecutor(ITypeRegistrar registrar)
...
     {
         if (configuration == null)
         {
             throw new ArgumentNullException(nameof(configuration));
-        }
+        }
+
+        args ??= new List<string>();
+
+
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -8,85 +8,97 @@ public CommandExecutor(ITypeRegistrar registrar)
...
+        _registrar.RegisterInstance(typeof(IConfiguration), configuration);
+        _registrar.RegisterLazy(typeof(IAnsiConsole), () => configuration.Settings.Console.GetConsole());
+
+
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -8,85 +8,97 @@ public CommandExecutor(ITypeRegistrar registrar)
...
-                {
-                    var console = configuration.Settings.Console.GetConsole();
-                    console.WriteLine(ResolveApplicationVersion(configuration));
-                    return 0;
-                }
-            }
-        }
+        _registrar.RegisterDependencies(model);
+
+
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -8,85 +8,97 @@ public CommandExecutor(ITypeRegistrar registrar)
...
+        // Asking for version? Kind of a hack, but it's alright.
+        // We should probably make this a bit better in the future.
+        if (args.Contains("-v") || args.Contains("--version"))
+        {
+            var console = configuration.Settings.Console.GetConsole();
+            console.WriteLine(ResolveApplicationVersion(configuration));
+            return 0;
+        }
+
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -8,85 +8,97 @@ public CommandExecutor(ITypeRegistrar registrar)
...
-        {
-            // Display help for default command.
-            configuration.Settings.Console.SafeRender(HelpWriter.WriteCommand(model, leaf.Command, configuration.Settings.ShowOptionDefaultValues));
-            return 1;
-        }
-
-        // Register the arguments with the container.
+        var parsedResult = ParseCommandLineArguments(model, configuration.Settings, args);
+
+
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -8,85 +8,97 @@ public CommandExecutor(ITypeRegistrar registrar)
...
+        // Register the arguments with the container.
         _registrar.RegisterInstance(typeof(CommandTreeParserResult), parsedResult);
         _registrar.RegisterInstance(typeof(IRemainingArguments), parsedResult.Remaining);
+
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -8,85 +8,97 @@ public CommandExecutor(ITypeRegistrar registrar)
...
+            // Create the content.
             var context = new CommandContext(parsedResult.Remaining, leaf.Command.Name, leaf.Command.Data);
 
             // Execute the command tree.
             return await Execute(leaf, parsedResult.Tree, context, resolver, configuration).ConfigureAwait(false);
         }
-    }
+    }
+#pragma warning restore SA1507 // Code should not contain multiple blank lines in a row
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -8,85 +8,97 @@ public CommandExecutor(ITypeRegistrar registrar)
...
 
-    private CommandTreeParserResult? ParseCommandLineArguments(CommandModel model, CommandAppSettings settings, IEnumerable<string> args)
+#pragma warning disable CS8603 // Possible null reference return.
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandInfo.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandInfo.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandInfo.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandInfo.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandModel.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandModel.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandModel.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandModel.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandOption.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandParameter.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandParameter.cs
FrankRay78: Done

test/Spectre.Console.Cli.Tests/Data/Help/CustomHelpProvider.cs
FrankRay78: Done

examples/Cli/Help/Help.csproj
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandParameter.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandModel.cs
FrankRay78: Done

src/Spectre.Console.Cli/Internal/Modelling/CommandInfo.cs
FrankRay78: The other review comments asked for unqualifying the namespace, as we add it to the global usings.  So I expect what is being asked for here is a typo, and I've made the changes as per all the other review comments.

src/Spectre.Console.Cli/Internal/Modelling/CommandArgument.cs
FrankRay78: The other review comments asked for unqualifying the namespace, as we add it to the global usings.  So I expect what is being asked for here is a typo, and so I've made the change to remove the namespace as per the other review comments below.

docs/input/cli/command-help.md
```diff
@@ -0,0 +1,47 @@
...
+
+The automatically generated help is derived from the configured commands and their command settings.
+
+The help is also context aware and tailored depending on what has been specified on the command line before it. For example,
+
+1. When `-h` or `--help` appears immediately after the application name (eg. `application.exe --help`), then the help displayed is a high-level summary of the application, including any command line examples and a listing of all possible commands the user can execute. 
+
+2. When `-h` or `--help` appears immediately after a command has been specified (eg. `application.exe command --help`), then the help displayed is specific to the command and includes information about command specific switches and any default values. 
+
+`Help.DefaultHelpProvider` is the `Spectre.Console` class responsible for determining context and preparing the help text to write to the console. It is an implementation of the public interface `Help.IHelpProvider`.
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Help/IHelpProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+/// `-h` or `--help` has been specified on the command line.
+/// </remarks>
+public interface IHelpProvider
+{
+    /// <summary>
+    /// Writes help information for the application.
+    /// </summary>
+    /// <param name="model">The command model to write help for.</param>
+    /// <returns>An enumerable collection of <see cref="IRenderable"/> objects representing the help information.</returns>
+    IEnumerable<IRenderable> Write(ICommandModel model);
```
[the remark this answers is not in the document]
FrankRay78: No it's not necessary, as is a hangover from the previous help writer implementation.  I've removed it now.

src/Spectre.Console.Cli/Help/ICommandContainer.cs
```diff
@@ -0,0 +1,25 @@
...
+
+/// <summary>
+/// Represents a command container.
+/// </summary>
+public interface ICommandContainer
+{
+    /// <summary>
+    /// Gets all the examples for the container.
+    /// </summary>
+    IList<string[]> Examples { get; }
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Help/ICommandContainer.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    /// <summary>
+    /// Gets all commands in the container.
+    /// </summary>
+    IList<ICommandInfo> Commands { get; }
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Help/ICommandInfo.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    /// <summary>
+    /// Gets a value indicating whether the command is hidden.
+    /// </summary>
+    bool IsHidden { get; }
+
+    /// <summary>
+    /// Gets the parameters associated with the command.
+    /// </summary>
+    IList<ICommandParameter> Parameters { get; }
```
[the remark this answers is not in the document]
FrankRay78: Done

src/Spectre.Console.Cli/Help/DefaultHelpProvider.cs
```diff
@@ -1,7 +1,28 @@
...
-{
+namespace Spectre.Console.Cli.Help;
+
+/// <summary>
+/// The default help provider for spectre.console.
+/// </summary>
+/// <remarks>
+/// Other IHelpProvider implementations can be injected into the CommandApp, if desired.
+/// </remarks>
+public class DefaultHelpProvider : IHelpProvider
```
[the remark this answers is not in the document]
FrankRay78: Agreed, done.

FrankRay78 (commented):
I've fully addressed all reviewer feedback provided.

patriksvensson:
@FrankRay78 Adding `global using Spectre.Console.Cli.Help` and removing the Help qualifier would work since the help stuff only is needed when implementing a help provider.

FrankRay78:
Hi @patriksvensson , ok cool.  This PR is complete and ready to merge then.

patriksvensson:
@FrankRay78 Awesome will take a look at it tonight or tomorrow 👍

patriksvensson (approved):
LGTM

PR #1262, opened 2023-07-23
Progress bar header and footer

### LINKED ISSUE #293: Title for the progress bar
First, very much thank you for this amazing console! It looks and works really cool!

I am working on the load testing tool NBomber and during the test run, I want to display(and dynamically update) some information over the progress bar.

I want to have the ability to set Title over the progress bar that I can change dynamically.
Here is an example of this functionality: https://github.com/Mpdreamz/shellprogressbar

I tried to use columns for this but it doesn't fully solve my issue. One of the problems is rendering the position of the progress bar - it changes based on text length.

phil-scott-78:
Allows a list of `IRenderable`s to be rendered for a header and footer for a progress bar. A user defined function is called each render tick with a list of the current tasks. A `Rows` collection is returned with the content to be included with the progress.

Would close #293

### DIFF
--- a/examples/Console/Progress/Program.cs
+++ b/examples/Console/Progress/Program.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using Spectre.Console;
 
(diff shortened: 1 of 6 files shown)


phil-scott-78:

phil-scott-78:
Another pattern that could be used would be a single `Func<IReadOnlyList<ProgressTask>, IRenderable,  IRenderable>` that takes in the completed output and allows the user to intercept it and wrap the whole thing with whatever they want. That would allow someone to add stuff horizontally in addition to vertically.

Same pattern could be used with the `ListPrompt` to allow someone to dump extra info regarding the currently selected item out.

nils-a (approved):
LGTM!

PR #1264, opened 2023-07-24
Add CommandExceptionHandler

JKamsker:
Reason:
I need a way of handling exceptions while having the typeresolver around. (To log to an implementation of ILogger for e.g)
Unfortunately, the existing exceptionhandler does not provide a way of accessing the ITypeResolver

Usage:

```csharp
public class GenericExceptionHandler : ICommandExceptionHandler
{
    // You can inject services here
    public GenericExceptionHandler()
    {
        
    }

        public bool Handle(CommandExceptionArgs args)
        {
            AnsiConsole.MarkupLine($"[red]{args.Exception.Message}[/]");
            return true;
        }
}

public class CommandGenericExceptionHandler : ICommandExceptionHandler<MyCommand>
{
    // You can inject services here
    public CommandGenericExceptionHandler()
    {
        
    }

        public bool Handle(CommandExceptionArgs args)
        {
            AnsiConsole.MarkupLine($"[red]{args.Exception.Message}[/]");
            return true;
        }
}
```

Then register them during startup: 
```
//...
services.AddTransient<ICommandExceptionHandler, GenericExceptionHandler>();
services.AddTransient< ICommandExceptionHandler<MyCommand>, CommandGenericExceptionHandler>();

//...
var app = new CommandApp(new TypeRegistrar(services));
```

### DIFF
--- a/src/Spectre.Console.Cli/CommandExceptionArgs.cs
+++ b/src/Spectre.Console.Cli/CommandExceptionArgs.cs
@@ -1,0 +1,30 @@
+namespace Spectre.Console.Cli;
+#pragma warning restore S2326
+
+/// <summary>
+/// Represents the args for a command exception.
+/// </summary>
+public readonly struct CommandExceptionArgs
+{
+    /// <summary>
+    /// Gets the command context.
+    /// </summary>
+    public CommandContext Context { get; }
+
+    /// <summary>
+    /// Gets the exception that was thrown.
+    /// </summary>
+    public Exception Exception { get; }
+
+    /// <summary>
+    /// Gets the command type.
+    /// </summary>
+    public Type? CommandType { get; }
+
+    internal CommandExceptionArgs(CommandContext context, Exception exception, Type? commandType)
+    {
+        Context = context;
+        Exception = exception;
+        CommandType = commandType;
+    }
+}
--- a/src/Spectre.Console.Cli/ICommandExceptionHandler.cs
+++ b/src/Spectre.Console.Cli/ICommandExceptionHandler.cs
@@ -1,0 +1,26 @@
+namespace Spectre.Console.Cli;
+
+/// <summary>
+/// Represents an args handler.
+/// Exception handlers are used to handle exceptions that occur during command execution.
+/// </summary>
+public interface ICommandExceptionHandler
+{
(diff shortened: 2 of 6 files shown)


nils-a:
Wouldn't it be easier to add the current `ITypeResolver` to the existing ErrorHandling?

JKamsker:
@nils-a I apologize for the delay. I fell ill over the weekend and am still recovering. I'll address this pull request once I'm in better shape.

JKamsker:
@nils-a Apologies, I'm having a bit of difficulty understanding what you're suggesting. Could you please explain it?

OT: Force pushed because i wanted a cleaner diff. The CommandExecutor appearantly is in mixed mode (LF & CRLF). vscode messes it up.

nils-a:
@JKamsker the currently [existing way to handle exceptions](https://spectreconsole.net/cli/exceptions) is:

```cs
var app = new CommandApp<DefaultCommand>();
app.Configure(config =>
{
    config.SetExceptionHandler(err =>
    {
        AnsiConsole.WriteException(err);
        // or whatever....
    });
});
```

I don't quite like the idea of adding another, different way to do exception handling. 

So: If you need the `ITypeResolver` while handling exceptions, my proposal was to add the `ITypeResolver` to the exception handler like this:

```cs
var app = new CommandApp<DefaultCommand>();
app.Configure(config =>
{
    config.SetExceptionHandler((err, resolver) =>
    {
        var console = (IAnsiConsole)resolver.Resolve(typeof(IAnsiConsole));
        console.WriteException(err);
        // or whatever....
    });
});
```

JKamsker:
@nils-a Oh yea, that seems like a neat idea! Why haven't i thought about this? 😅 
Would adding an overload like ``SetExceptionHandler<TCommand>`` be acceptable?

nils-a:
> Would adding an overload like SetExceptionHandler<TCommand> be acceptable?

Hm. Do you really need different handlers for different Commands? I have the feeling this would make things rather complicated. 
Wouldn't it be easier, in that case, to create your own `ICommand` that is capable of handling exceptions?

FrankRay78:
Superseded by https://github.com/spectreconsole/spectre.console/pull/1411.

Issue #1258, opened 2023-07-18
Extendibility of elements


dvdking:
The problem is modifying internal behaviour of an element without writing a completely new one.

Let's take as an example SelectionPrompt. I wanted to add additional input handling to handle numbers from input as a selection index.  Right now, there is no way to do that, as the class itself is sealed, and all inner classes are internal, so I cannot even create a simple copy of the class but need to copypaste the whole hierarchy of internal dependencies or create my own from ground up or not use nuget package and have copy of lib code.
Other places also might benefit from extendibility, for example customization of rendering.

Possible solution is to avoid sealed internal classes and non-virtual methods without possibility to override their behaviour.
Alternatively leave classes sealed but all internals could be made public to create custom classes.
And the third, the safest, option is to add hooks to the main functionality of elements. Like it is done with SelectionPrompt.Converter.

PR #1289, opened 2023-09-03
Selection Prompt Search

slang25:
Multi Selection Search

This PR adds a couple of features to Selection Prompt:
* Search as you type
* Skipping selection of group nodes for `Leaf` selection mode

I've yet to add the tests which is a bit naughty, but wanted to share what I have so far to start some discussion. Is this feature welcome, or is there a better direction?

### DIFF
--- a/src/Spectre.Console/Extensions/StringExtensions.cs
+++ b/src/Spectre.Console/Extensions/StringExtensions.cs
@@ -187,6 +187,13 @@ public static class StringExtensions
 #endif
     }
 
+#if NETSTANDARD2_0
+    internal static bool Contains(this string target, string value, System.StringComparison comparisonType)
+    {
+        return target.IndexOf(value, comparisonType) != -1;
+    }
+#endif
+
     /// <summary>
     /// "Masks" every character in a string.
     /// </summary>
--- a/src/Spectre.Console/Extensions/StringExtensions.cs
+++ b/src/Spectre.Console/Extensions/StringExtensions.cs
@@ -202,18 +202,11 @@ public static class StringExtensions
     /// <returns>Masked string.</returns>
     public static string Mask(this string value, char? mask)
     {
-        var output = string.Empty;
-
         if (mask is null)
         {
-            return output;
+            return string.Empty;
         }
 
-        foreach (var c in value)
-        {
-            output += mask;
-        }
-
-        return output;
+        return new string(mask.Value, value.Length);
     }
 }
--- a/src/Spectre.Console/Extensions/StringExtensions.cs
+++ b/src/Spectre.Console/Extensions/StringExtensions.cs
@@ -209,4 +209,98 @@ public static class StringExtensions
 
         return new string(mask.Value, value.Length);
     }
+
+    /// <summary>
+    /// Highlights the first text match in provided value.
+    /// </summary>
+    /// <param name="value">Input value.</param>
+    /// <param name="searchText">Text to search for.</param>
+    /// <param name="highlightStyle">The style to apply to the matched text.</param>
+    /// <returns>Markup of input with the first matched text highlighted.</returns>
+    internal static string Highlight(this string value, string searchText, Style? highlightStyle)
+    {
+        if (value is null)
+        {
+            throw new ArgumentNullException(nameof(value));
+        }
+
+        if (searchText is null)
+        {
+            throw new ArgumentNullException(nameof(searchText));
+        }
+
+        if (highlightStyle is null)
+        {
+            throw new ArgumentNullException(nameof(highlightStyle));
+        }
+
+        if (searchText.Length == 0)
+        {
+            return value;
+        }
+
+        var foundSearchPattern = false;
+        var builder = new StringBuilder();
+        using var tokenizer = new MarkupTokenizer(value);
+        while (tokenizer.MoveNext())
+        {
+            var token = tokenizer.Current!;
+
+            switch (token.Kind)
+            {
+                case MarkupTokenKind.Text:
+                    {
+                        var tokenValue = token.Value;
+                        if (tokenValue.Length == 0)
+                        {
+                            break;
+                        }
+
+                        if (foundSearchPattern)
+                        {
+                            builder.Append(tokenValue);
+                            break;
+                        }
+
+                        var index = tokenValue.IndexOf(searchText, StringComparison.OrdinalIgnoreCase);
+                        if (index == -1)
+                        {
+                            builder.Append(tokenValue);
+                            break;
+                        }
+
+                        foundSearchPattern = true;
+                        var before = tokenValue.Substring(0, index);
+                        var match = tokenValue.Substring(index, searchText.Length);
+                        var after = tokenValue.Substring(index + searchText.Length);
+
+                        builder
+                            .Append(before)
+                            .AppendWithStyle(highlightStyle, match)
+                            .Append(after);
+
+                        break;
+                    }
+
+                case MarkupTokenKind.Open:
+                    {
+                        builder.Append("[" + token.Value + "]");
+                        break;
+                    }
+
+                case MarkupTokenKind.Close:
+                    {
+                        builder.Append("[/]");
+                        break;
+                    }
+
+                default:
+                    {
+                        throw new InvalidOperationException("Unknown markup token kind.");
+                    }
+            }
+        }
+
+        return builder.ToString();
+    }
 }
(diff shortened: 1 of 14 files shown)


slang25:
This is somewhat related to #1258, as this is a feature I'd really like in my CLI, however my only feasible options are to fork the library or hope I can get this merged.

src/Spectre.Console/Extensions/StringExtensions.cs
```diff
@@ -185,28 +185,35 @@ internal static bool ContainsExact(this string text, string value)
...
+        {
+            return output;
+        }
+
+        foreach (var c in value)
+        {
+            output += mask;
+        }
+
+        return output;
```
martincostello, line 217: Could you make this just:

```csharp
if (mask is null)
{
    return string.Empty;
}

return new string(mask, value.Length);
```

src/Spectre.Console/Prompts/SelectionPrompt.cs
```diff
@@ -53,6 +58,11 @@ public sealed class SelectionPrompt<T> : IPrompt<T>, IListPromptStrategy<T>
     /// </summary>
     public SelectionMode Mode { get; set; } = SelectionMode.Leaf;
 
+    /// <summary>
+    /// Gets or sets a value indicating whether or not the search filter is enabled.
+    /// </summary>
+    public bool SearchFilterEnabled { get; set; } = false;
```
martincostello, line 64: As it's already the default:

```suggestion
    public bool SearchFilterEnabled { get; set; }
```

GOATS2K:
This looks incredibly helpful when managing selection prompts with tons of options - I'd love to see this merged soon.

slang25:
Thanks @GOATS2K, I'm pretty happy with it and am using it in one of my apps. I've just updated the API names to be a bit more sensible.

@patriksvensson what are your thoughts on this?

src/Spectre.Console/Prompts/SelectionPrompt.cs
```diff
@@ -53,6 +58,11 @@ public sealed class SelectionPrompt<T> : IPrompt<T>, IListPromptStrategy<T>
     /// </summary>
     public SelectionMode Mode { get; set; } = SelectionMode.Leaf;
 
+    /// <summary>
+    /// Gets or sets a value indicating whether or not search is enabled.
+    /// </summary>
+    public bool SearchEnabled { get; set; }
```
patriksvensson, line 64: We should also add a setting for the text displayed at the bottom of the list `(Type to search)`

src/Spectre.Console/Prompts/List/IListPromptStrategy.cs
```diff
@@ -7,6 +7,11 @@ namespace Spectre.Console;
 internal interface IListPromptStrategy<T>
     where T : notnull
 {
+    /// <summary>
+    /// Gets a value indicating whether or not the prompt should skip unselectable items.
+    /// </summary>
+    public bool ShouldSkipUnselectableItems { get; }
```
patriksvensson, line 13: Can't this be a parameter for `ListPrompt<T>.Show` instead?

src/Spectre.Console/Prompts/SelectionPrompt.cs
```diff
@@ -93,7 +106,7 @@ public async Task<T> ShowAsync(IAnsiConsole console, CancellationToken cancellat
     /// <inheritdoc/>
     ListPromptInputResult IListPromptStrategy<T>.HandleInput(ConsoleKeyInfo key, ListPromptState<T> state)
     {
-        if (key.Key == ConsoleKey.Enter || key.Key == ConsoleKey.Spacebar || key.Key == ConsoleKey.Packet)
+        if (key.Key == ConsoleKey.Enter || key.Key == ConsoleKey.Packet)
```
patriksvensson, line 109: Why is `ConsoleKey.Spacebar` removed here?

patriksvensson (changes requested):
This looks good, but I think some minor things need to be addressed. I would also like to see some tests for the search functionality.

src/Spectre.Console/Prompts/SelectionPrompt.cs
slang25: This was from testing, I was entertaining the idea of making it configurable. You'd want spacebar to select a multi-selection prompt, but for a regular selection prompt you might want to search for multiple words.

I'll revert it 👍

slang25:
@patriksvensson thanks for taking a look, I've addressed the feedback. I've added some tests, I could add some Verify tests too and cover more scenarios but am on the fence, let me know what you think.

src/Spectre.Console/Prompts/SelectionPrompt.cs
```diff
@@ -169,15 +195,43 @@ IRenderable IListPromptStrategy<T>.Render(IAnsiConsole console, bool scrollable,
                 text = text.RemoveMarkup().EscapeMarkup();
             }
 
+            if (searchText.Length > 0 && !(item.Node.IsGroup && Mode == SelectionMode.Leaf))
+            {
+                var index = text.IndexOf(searchText, StringComparison.OrdinalIgnoreCase);
```
sleepyfran, line 200: This seems to match against markup in the text and causes problems when it matches against it. I took this for a spin on my project and when the line is not current (since that automatically escapes it) it inserts the styles of the search in the middle of other markup, which ends up causing an exception when parsing the markup:
image

The original text is `The Hobbit ([bold green3]15.00d$[/])`, if I search for `l` then it matches with the `l` in `bold` (which is markup) and ends up creating `The Hobbit ([bo[bold on yellow]l[/]d green3]15.00d$[/])`, which is invalid markup since styles cannot be nested.

Escaping it seems to work fine, though, so I guess expanding the if in line 193 for search makes sense 😄

sleepyfran (commented):
Thanks a lot for this, really excited to use it! I took it for a spin in a personal fork and found the issue I commented about. Other than that it seems to work perfectly 🙇🏻

src/Spectre.Console/Prompts/SelectionPrompt.cs
slang25: Thanks, this is a nice find and I'm looking at fixing it now. It's simple enough to discard the markup during the search, but I think the correct behaviour would be to override the style if the found text is within a style (or perhaps in the edge case; spanning across multiple styles), which will be a bit more involved.

Just having a think about it now.

slang25: I wrote something that would highlight matches across different markup tokens and it was very convoluted and had lots of edge cases. This led me to realise that it's probably best to say that we don't search across tokens.

I've pushed an update that is much simpler, please try again when you get some time 🙂

slang25:
Hi @patriksvensson, please could you take another look at this PR when you get some time. I'm happy to address any further feedback 🙂

Jacob-32587:
Would love to see this go through soon, extremely useful feature when dealing with large prompts.

patriksvensson:
@slang25 I'm rebasing this and fixing the merge conflicts right now

patriksvensson:

slang25:
Thanks @patriksvensson, this library is such a gem for the .NET community, I hope you have time to keep it going 🙂  I was worried we'd lose you to swift 😅

patriksvensson:
@slang25 Haha, thanks! I still have a lot of love for dotnet 🙂

fihorvat:
@slang25 Sorry for the late comment but it looks like there is a bug. If you try to use something other than `string` for `SelectionPrompt<T>` in combination with the `UseConverter` the cursor selection doesn't work how it should.
Text highlighting works but the cursor doesn't move.
 
video

Here is a simple test data:

```c#
class TestItem
{
  public string Name { get; set; }
  public string Url { get; set; }
}

var items = new TestItem[] {
  new() { Name = "banana", Url = "something" },
  new() { Name = "orange", Url = "something" },
  new() { Name = "apple", Url = "something" }
};

var prompt = new SelectionPrompt<TestItem>()
  .Title("Select [green]fruit[/]?")
  .PageSize(10)
  .AddChoices(items)
  .UseConverter(x => x.Name)
  .EnableSearch();
prompt.SearchHighlightStyle(Style.Parse("yellow1"));
var response = AnsiConsole.Prompt(prompt);
```

And could you please add the method `SearchHighlightStyle` same as we have for `HighlightStyle`?

patriksvensson:
@fihorvat This PR had been merged and no further work is expected by the contributor. Please open a new issue if you experience any bugs.

Issue #1279, opened 2023-08-19
Missing columns GIF/preview assets on the documentation


MangelMaxime:
On the [Columns](https://spectreconsole.net/widgets/columns) documentation page, the GIF/preview at the page doesn't load.

<img width="971" alt="CleanShot 2023-08-19 at 21 19 34@2x" src="https://github.com/spectreconsole/spectre.console/assets/4760796/cc9664c3-ca5c-4a0a-a16d-3fa20c793ebf">

Looking at the console I get this error:

<img width="475" alt="CleanShot 2023-08-19 at 21 19 08@2x" src="https://github.com/spectreconsole/spectre.console/assets/4760796/18bd6792-95b9-4bf2-850f-ef85bd640aa1">

PR #1295, opened 2023-09-12
Update ColumnsSample to showcase nicer data

nils-a:
Also, I switched the base width of the AsciiCast back to 82 (from 120) so newly created casts are no longer overflowing the display with on the pages.

Re-created the Panel and BreakdownChart casts, as
they were currently overflowing due to the 120 chars width of the cast.


nils-a:
This PR is still roughly related to https://github.com/spectreconsole/spectre.console/issues/1279

patriksvensson (approved):
LGTM 👍

Issue #1189, opened 2023-03-04
Using centered figlet in rows throws


EluciusFTW:
**Information**
 - OS: Windows 10
 - Version: 0.46
 - Terminal: Windows Terminal

**Describe the bug**
Using a centered figlet inside of rows leads to an exception, when there is another 'small element' present.

**To Reproduce**
Here is a minimal console app to reproduce the bug:
```cs
using Spectre.Console;

var rows = new Rows(
    new Markup("Short"), // omitting this markup altogether or making the string leads to no exception
    new FigletText("Some other string").Centered()); // using left- or right-justified leads to no exception in any case
var panel = new Panel(rows);
AnsiConsole.Write(panel);
```
Note: Either removing the first `IRenderable`, or making its content bigger by passign a longer string will make it **not throw**.

Here's the StackTrace:

```
Unhandled exception. System.ArgumentOutOfRangeException: Count cannot be less than zero. (Parameter 'count')
   at System.String.Ctor(Char c, Int32 count)
   at Spectre.Console.Rendering.Segment.Padding(Int32 size) in /_/src/Spectre.Console/Rendering/Segment.cs:line 53
   at Spectre.Console.FigletText.Render(RenderOptions options, Int32 maxWidth)+MoveNext() in /_/src/Spectre.Console/Widgets/Figlet/Figl
etText.cs:line 73
   at Spectre.Console.EnumerableExtensions.Enumerate[T](IEnumerator`1 source)+MoveNext() in /_/src/Spectre.Console/Internal/Extensions/
EnumerableExtensions.cs:line 102
   at Spectre.Console.Rows.Render(RenderOptions options, Int32 maxWidth) in /_/src/Spectre.Console/Widgets/Rows.cs:line 55
   at Spectre.Console.Rendering.Renderable.Spectre.Console.Rendering.IRenderable.Render(RenderOptions options, Int32 maxWidth) in /_/sr
c/Spectre.Console/Rendering/Renderable.cs:line 19
   at Spectre.Console.Padder.Render(RenderOptions options, Int32 maxWidth) in /_/src/Spectre.Console/Widgets/Padder.cs:line 69
   at Spectre.Console.Rendering.Renderable.Spectre.Console.Rendering.IRenderable.Render(RenderOptions options, Int32 maxWidth) in /_/sr
c/Spectre.Console/Rendering/Renderable.cs:line 19
[...]

PR #1302, opened 2023-09-14
Fix figlet centering possibly throwing due to negative size

olabacker:
Fixing this issue https://github.com/spectreconsole/spectre.console/issues/1189

Zeroing values which can end up negative causing Segment.Padding ctor to throw


patriksvensson (approved):
LGTM 👍

PR #1349, opened 2023-10-30
HelpProvider supports localisation

FrankRay78:
+ Extracted hardcoded strings into a default resource file
+ Included localised strings for French
+ Wrote unit tests for both

image


FrankRay78:
@patriksvensson - I've removed the need to set the culture on the current thread by eliminating statics, and so ensured that no funny business will happen when parallel running tests.

test/Spectre.Console.Cli.Tests/Unit/CommandAppTests.Help.cs
```diff
@@ -250,6 +252,62 @@ public Task Should_Output_Default_Command_And_Additional_Commands_When_Default_C
...
+                configurator.AddCommand<GiraffeCommand>("giraffe");
+            });
+
+            // When
+            var result = fixture.Run();
+
+            // Then
+            var settings = new VerifySettings();
+            settings.DisableRequireUniquePrefix();
+            return Verifier.Verify(result.Output, settings);
```
nils-a, line 308: I think it should be possible to re-use the `EN` method like this:

```suggestion
           return  Should_Output_Default_Command_And_Additional_Commands_When_Default_Command_Has_Required_Parameters_And_Is_Called_Without_Args_EN(culture);
```

That would spare me copying the content for a third time when I add the German translation.

test/Spectre.Console.Cli.Tests/Unit/CommandAppTests.Help.cs
```diff
@@ -1,3 +1,4 @@
+using System.Diagnostics.SymbolStore;
```
nils-a, line 1: I think this line is not needed.

nils-a (approved):
Two minor changes, but generally LGTM!

src/Spectre.Console.Cli/ConfiguratorExtensions.cs
```diff
@@ -39,7 +39,24 @@ public static IConfigurator SetHelpProvider<T>(this IConfigurator configurator)
...
-    }
+    }
+
+    /// <summary>
+    /// Sets the culture for the application.
+    /// </summary>
+    /// <param name="configurator">The configurator.</param>
+    /// <param name="culture">The name of the culture.</param>
+    /// <returns>A configurator that can be used to configure the application further.</returns>
+    public static IConfigurator SetApplicationCulture(this IConfigurator configurator, string culture)
```
patriksvensson, line 50: Perhaps we could use a `CultureInfo` here instead and take the culture from that instance.

That way we won't need to document what a valid culture name is or is not.

patriksvensson (changes requested):
A minor thing that I think we should dicuss.

test/Spectre.Console.Cli.Tests/Unit/CommandAppTests.Help.cs
FrankRay78: Agreed.

test/Spectre.Console.Cli.Tests/Unit/CommandAppTests.Help.cs
FrankRay78: You mean collapsing both of these tests together @nils-a ?

```csharp
        // Localised version of the above test, in English
        [Theory]
        [InlineData(null)]
        [InlineData("")]
        [InlineData("en")]
        [InlineData("en-EN")]
        [Expectation("Default_Without_Args_Additional")]
        public Task Should_Output_Default_Command_And_Additional_Commands_When_Default_Command_Has_Required_Parameters_And_Is_Called_Without_Args_EN(string culture)
```

and

```csharp
        // Localised version of the above test, in French
        [Theory]
        [InlineData("fr")]
        [InlineData("fr-FR")]
        [Expectation("Default_Without_Args_Additional-FR")]
        public Task Should_Output_Default_Command_And_Additional_Commands_When_Default_Command_Has_Required_Parameters_And_Is_Called_Without_Args_FR(string culture)
```

I think an elegant solution might be to dynamically vary the expectation by appending the culture name to something like "Default_Without_Args_Additional-"

However, the following cases:

        [InlineData(null)]
        [InlineData("")]

Might actually need to become something like:

        [InlineData(null, "en")]
        [InlineData("", "en")]

In order to qualify the localisation result we expect ie. qualifying the expectation.

FrankRay78:
@patriksvensson , @nils-a

FYI. If you provide me with the localised strings for the following in your own language, along with the culture, I can see about pushing these into this branch ahead of merging:

Prints help information
Prints version information
DESCRIPTION
USAGE
OPTIONS
COMMAND
COMMANDS
EXAMPLES
ARGUMENTS
DEFAULT

patriksvensson:
Sure!

**Prints help information**
Skriver ut hjälpinformatiob
 **Prints version information**
Skriver ut versionsnummer
 **DESCRIPTION**
BESKRIVNING
 **USAGE**
ANVÄNDING
 **OPTIONS**
VAL
 **COMMAND**
KOMMANDO
 **COMMANDS**
KOMMANDON
 **EXAMPLES**
EXEMPEL
 **ARGUMENTS**
ARGUMENT
 **DEFAULT**
STANDARD

nils-a:
en | de
--|--
Prints help information | Zeigt Hilfe an
Prints version information | Zeigt Versionsinformationen an
DESCRIPTION | BESCHREIBUNG
USAGE | VERWENDUNG
OPTIONS | OPTIONEN
COMMAND | KOMMANDO
COMMANDS | KOMMANDOS
EXAMPLES | BEISPIELE
ARGUMENTS | ARGUMENTE
DEFAULT | STANDARDWERT

0xced:
For French I'd go _slightly_ less verbose.

en | fr
--|--
Prints help information | _Affiche l'aide_
Prints version information | _Affiche la version_
USAGE | UTILISATION (was missing in the screenshot)

src/Spectre.Console.Cli/ICommandAppSettings.cs
```diff
@@ -5,6 +5,11 @@ namespace Spectre.Console.Cli;
 /// </summary>
 public interface ICommandAppSettings
 {
+    /// <summary>
+    /// Gets or sets the culture.
```
0xced, line 9: We should document what culture is used by default (i.e. when `null`)

0xced: Or even change this property to a non nullable `CultureInfo` (as already mentioned by Patrick elsewhere) but document what's the default value anyway.

FrankRay78: I added the following:

```csharp
    /// <summary>
    /// Gets or sets the culture.
    /// </summary>
    /// <remarks>
    /// Text displayed by <see cref="Help.HelpProvider"/> can be localised, but defaults to English.
    /// Setting this property informs the resource manager which culture to use when fetching strings.
    /// English will be used when a culture has not been specified (ie. this property is null)
    /// or a string has not been localised for the specified culture.
    /// </remarks>
    CultureInfo? Culture { get; set; }
```

Do you think it sufficiently documents the behaviour @0xced ?

FrankRay78:
@patriksvensson, @nils-a, @0xced - I believe the latest push addresses your excellent review comments, as well as updating the French translations (as suggested) and adding Swedish and German localisations.

FrankRay78 (commented):
I'm happy with the changes.

patriksvensson (approved):

src/Spectre.Console.Cli/ICommandAppSettings.cs
0xced: Yes, that's very good! I think this remark should also be added to the `ConfiguratorExtensions.SetApplicationCulture` method because currently, _Sets the culture for the application._ could be confusing.

src/Spectre.Console.Cli/Resources/HelpProvider.fr.resx
```diff
@@ -0,0 +1,141 @@
...
+  <resheader name="version">
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="Command" xml:space="preserve">
```
0xced, line 120: I think `ARGUMENTS`, `DESCRIPTION` and `OPTIONS` should be added to the French translation, even if they are the exact same words. This makes it clear that the localisation was not forgotten. Also, Rider emits warnings saying that it's missing French translations for those 3 elements.

FrankRay78: I will make these changes later today @0xced, then you will be free to review and merge.

test/Spectre.Console.Cli.Tests/Spectre.Console.Cli.Tests.csproj
```diff
@@ -30,9 +30,8 @@
   </ItemGroup>
 
   <ItemGroup>
```
nils-a, line 32: I'm nit-picking here, but, looking at it now, I feel this whole `ItemGroup` is not really needed.

FrankRay78: note to self: Never try and sneak anything past @nils-a , he will find out.

FrankRay78:
@0xced Addressed all further review comments in the latest push.

src/Spectre.Console.Cli/ConfiguratorExtensions.cs
```diff
@@ -39,7 +39,30 @@ public static IConfigurator SetHelpProvider<T>(this IConfigurator configurator)
...
+    /// <param name="configurator">The configurator.</param>
+    /// <param name="culture">The culture.</param>
+    /// <returns>A configurator that can be used to configure the application further.</returns>
+    /// <remarks>
+    /// Text displayed by <see cref="Help.HelpProvider"/> can be localised, but defaults to English.
+    /// Setting the application culture informs the resource manager which culture to use when fetching strings.
+    /// English will be used when a culture has not been specified
+    /// or a string has not been localised for the specified culture.
+    /// </remarks>
+    public static IConfigurator SetApplicationCulture(this IConfigurator configurator, CultureInfo culture)
```
0xced, line 56: I think we should either explicitly allow a `null` culture (by changing `CultureInfo` into `CultureInfo?` and adapting the documentation of the `culture` parameter) or throw an `ArgumentNullException`. Currently, the test code emits a warning (_Possible 'null' assignment to non-nullable entity_) but actually executes fine, even when `null` is passed:

```csharp
configurator.SetApplicationCulture(string.IsNullOrEmpty(culture) ? null : new CultureInfo(culture))
```

I'm not sure which option is best. I'm trying to think about how it's going to be used at the application level, probably something like this:
```csharp
if (userPrefersLocalizedCli)
{
    configurator.SetApplicationCulture(CultureInfo.CurrentUICulture)
}
```

This use case would imply throwing an `ArgumentNullException` but my gut feeling says `CultureInfo?` is fine so I'm happy to hear your opinions on this.

Thinking more about it, `userPrefersLocalizedCli` might be quite tricky to compute so it might end up being the developer's choice to unconditionally call `SetApplicationCulture(CultureInfo.CurrentUICulture)`. This raises the question if Spectre.Console.Cli should default to `CultureInfo.CurrentUICulture` instead of English.

Sorry for being so nitpicky about this pull request, Frank. 🫣

0xced: After pondering two more minutes, I think we should throw `ArgumentNullException` because it doesn't make sense to call `SetApplicationCulture` with `null` since it's already the default. If you don't want to change the default then just don't call `SetApplicationCulture`. Seems sensible?

0xced: And I think the corresponding tests could be merged into one single (slightly adapted) test.

```csharp
        [Theory]
        [InlineData(null, "EN")]
        [InlineData("iv", "EN")]
        [InlineData("en", "EN")]
        [InlineData("en-EN", "EN")]
        [InlineData("fr", "FR")]
        [InlineData("fr-FR", "FR")]
        [InlineData("sv", "SV")]
        [InlineData("sv-SE", "SV")]
        [InlineData("de", "DE")]
        [InlineData("de-DE", "DE")]
        [Expectation("Default_Without_Args_Additional")]
        public Task Should_Output_Default_Command_And_Additional_Commands_When_Default_Command_Has_Required_Parameters_And_Is_Called_Without_Args(string culture, string expectationPrefix)
        {
            // Given
            var fixture = new CommandAppTester();
            fixture.SetDefaultCommand<LionCommand>();
            fixture.Configure(configurator =>
            {
                configurator.AddExample("20", "--alive");
                configurator.SetApplicationName("myapp");
                configurator.AddCommand<GiraffeCommand>("giraffe");
                if (culture != null)
                {
                    configurator.SetApplicationCulture(CultureInfo.GetCultureInfo(culture));
                }
            });

            // When
            var result = fixture.Run();

            // Then
            return Verifier.Verify(result.Output).UseTextForParameters(expectationPrefix).DisableRequireUniquePrefix();
        }
```

And the `Default_Without_Args_Additional.Output.verified.txt` file can be deleted.

FrankRay78: > After pondering two more minutes, I think we should throw `ArgumentNullException` because it doesn't make sense to call `SetApplicationCulture` with `null` since it's already the default. If you don't want to change the default then just don't call `SetApplicationCulture`. Seems sensible?

I was of the same opinion, and came to say as such.  But you beat me to it.  I don't mind the clarifications as we want to get it right.  

As an aside, I think Patrik was right to suggest the user sets the culture explicitly, rather than embed a possible call to fetch the Culture from the UI thread - ie. put the responsibility back on the developer to define exactly how it should operate.  I'll make the changes and do another push.

patriksvensson: I think we should accept null here. Perhaps the user want to change it or set the value dynamically somehow. We shouldn't throw.

FrankRay78: Hi @patriksvensson, the fallback resource text is currently returned for a null culture. I guess what your suggestion does is allow the user to return to that state, if they wish to.

patriksvensson: Exactly, and it also makes it easier to set the value. Our other API:s in SC works this way.

Makes it possible to do:

```csharp
config.SetApplicationCulture(GetMyAppCulture())
```

Instead of:

```csharp
var culture = GetMyAppCulture();
if(culture != null)
{
    config.SetApplicationCulture(culture);
}
```

If the property is nullable, we should accept null for convenience methods.

0xced: It would actually be hard to write a `GetMyAppCulture()` method that returns `null` since both `CultureInfo.CurrentUICulture` and `CultureInfo.GetCultureInfo()` return a non null `CultureInfo` but let's accept the nullable version since it's easier to use.

Just yesterday, I was bothered by the TextPromptExtensions methods `DefaultValueStyle()` and `ChoicesStyle()` that don't accept nullable Style instances while the underlying properties are nullable. I can be inconsistent sometimes. 😆

patriksvensson: Using those methods sure, but imagine if someone is doing a database lookup for a culture that the user set somehow, or checking some other configuration in a configuration file.

FrankRay78: The latest push has made it nullable @patriksvensson, and also removed the redundant test @0xced .

patriksvensson (approved):

Tolitech:
Hello @patriksvensson,
I would like the HelpProvider to support some additional languages, as listed below.

image

Would it be possible for me to contribute by creating the Resource files that are not yet supported?

FrankRay78:
Yes please @Tolitech! I can review and merge.

Tolitech:
Hello, @FrankRay78.
The changes have already been made in my FORK.
I have created 6 new resource files.
I have tested it locally and will submit the Pull Request soon.

PR #1358, opened 2023-11-09
Fix AnsiConsoleOutput safe height

0xced:
The safe height (introduced in 3e2eea730bf0f6fe4a5e336faa312b526d351079) would be 80 (the value of the safe width) instead of 24.

Removing the explicit `defaultValue` parameter ensures that the correct constants are used, i.e. Constants.DefaultTerminalWidth and Constants.DefaultTerminalHeight.


patriksvensson (approved):
LGTM! 👍

Issue #1405, opened 2023-12-15
Support multiple interceptors together


aditnryn:
**Is your feature request related to a problem? Please describe.**
Add support for invoking two or more interceptors.

**Describe the solution you'd like**
Replace `SetInterceptor` on `IConfigurator` with `AddInterceptor` method that allows chaining of interceptors.

**Describe alternatives you've considered**
Today, one can create the class `ChainedInterceptor` that delegates to multiple interceptors to circumvent this feature gap.

```csharp
class ChainedInterceptor(params ICommandInterceptor[] interceptors) : ICommandInterceptor
{
    public void Intercept(CommandContext context, CommandSettings settings)
    {
        foreach (var interceptor in interceptors)
        {
            interceptor.Intercept(context, settings);
        }
    }
}
```

I'm interested in having library support for interceptor chaining.

**Additional context**

aditnryn:
I'll be happy to contribute code to close this feature request. However, I'm not sure how to achieve this feature request without causing backward-incompatible changes to either the `IConfigurator` or the `ICommandAppSettings` interface.

patriksvensson:
Wouldn't it be better if we allowed multiple interceptors to be registered instead of a single one?

aditnryn:
Yes, that's the intention of the feature request too. The `SetInterceptor` method can be deprecated in favour of an `AddInterceptor` method on the `IConfigurator` interface IMO.

nils-a:
@aditnryn would #1412 solve your problem?

FrankRay78:
Fixed by https://github.com/spectreconsole/spectre.console/pull/1412.

PR #1412, opened 2023-12-23
Add the possibility to register multiple interceptors

nils-a:
related to #1405 
related to #1093 

- [x] I have read the [Contribution Guidelines](../CONTRIBUTING.md)
- [x] A maintainer has signed off on the changes and the issue was assigned to me

## Changes

The registration of the interceptor is now handled in `ITypeRegistrar` this makes it possible to register multiple interceptors and also to use DI in interceptors.
Additionally, an `InterceptResult`-Method was introduced that runs after the command and can be used to alter the result value or to do some cleanup after the command.

### DIFF
--- a/docs/input/cli/commandApp.md
+++ b/docs/input/cli/commandApp.md
@@ -81,9 +81,12 @@ Hint: If you do write your own implementation of `TypeRegistrar` and `TypeResolv
 there is a utility `TypeRegistrarBaseTests` available that can be used to ensure your implementations adhere to the required implementation. Simply call `TypeRegistrarBaseTests.RunAllTests()` and expect no `TypeRegistrarBaseTests.TestFailedException` to be thrown.
 
 ## Interception
+Interceptors can be registered with the `TypeRegistrar` (or with a custom DI-Container). Alternatively, `CommandApp` also provides a `SetInterceptor` configuration.
 
-`CommandApp` also provides a `SetInterceptor` configuration. An interceptor is run before all commands are executed. This is typically used for configuring logging or other infrastructure concerns.
+All interceptors must implement `ICommandInterceptor`. Upon execution of a command, The `Intercept`-Method of an instance of your interceptor will be called with the parsed settings. This provides an opportunity for configuring any infrastructure or modifying the settings.
+When the command has been run, the `InterceptResult`-Method of the same instance is called with the result of the command.
+This provides an opportunity to modify the result and also to tear down any infrastructure in use.
 
-All interceptors must implement `ICommandInterceptor`. Upon execution of a command, an instance of your interceptor will be called with the parsed settings. This provides an opportunity for configuring any infrastructure or modifying the settings.
+The `Intercept`-Method of each interceptor is run before the command is executed and the `InterceptResult`-Method is run after it. These are typically used for configuring logging or other infrastructure concerns.
 
 For an example of using the interceptor to configure logging, see the [Serilog demo](https://github.com/spectreconsole/spectre.console/tree/main/examples/Cli/Logging).
--- a/src/Spectre.Console.Cli/ConfiguratorExtensions.cs
+++ b/src/Spectre.Console.Cli/ConfiguratorExtensions.cs
@@ -229,7 +229,7 @@ public static class ConfiguratorExtensions
             throw new ArgumentNullException(nameof(configurator));
         }
 
-        configurator.Settings.Interceptor = interceptor;
+        configurator.Settings.Registrar.RegisterInstance<ICommandInterceptor>(interceptor);
         return configurator;
     }
 
--- a/src/Spectre.Console.Cli/ICommandAppSettings.cs
+++ b/src/Spectre.Console.Cli/ICommandAppSettings.cs
@@ -50,6 +50,7 @@ public interface ICommandAppSettings
     /// Gets or sets the <see cref="ICommandInterceptor"/> used
     /// to intercept settings before it's being sent to the command.
     /// </summary>
+    [Obsolete("Register the interceptor with the ITypeRegistrar.")]
     ICommandInterceptor? Interceptor { get; set; }
 
     /// <summary>
--- a/src/Spectre.Console.Cli/ICommandInterceptor.cs
+++ b/src/Spectre.Console.Cli/ICommandInterceptor.cs
@@ -12,5 +12,25 @@ public interface ICommandInterceptor
     /// </summary>
     /// <param name="context">The intercepted <see cref="CommandContext"/>.</param>
     /// <param name="settings">The intercepted <see cref="CommandSettings"/>.</param>
-    void Intercept(CommandContext context, CommandSettings settings);
+    void Intercept(CommandContext context, CommandSettings settings)
+#if NETSTANDARD2_0
+    ;
+#else
+    {
+    }
+#endif
+
+    /// <summary>
+    /// Intercepts a command result before it's passed as the result.
+    /// </summary>
+    /// <param name="context">The intercepted <see cref="CommandContext"/>.</param>
+    /// <param name="settings">The intercepted <see cref="CommandSettings"/>.</param>
+    /// <param name="result">The result from the command execution.</param>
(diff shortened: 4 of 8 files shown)


nils-a:
@spectreconsole/maintainers what do you think?

rcdailey:
@nils-a I took a look at the code changes. I even pulled down your branch on my machine, did a `dotnet pack` and `dotnet nuget push` to a local feed just to be able to use your changes in action.

My overall first impression is positive! The additions to `ICommandInterceptor` has allowed me to clean up a significant chunk of what I had in my `Program.cs`, especially regarding the "Hacks" I had to implement.

There's one small gap, however. When an exception occurs, I need to be able to handle that *with access to DI*. Unfortunately, the `InterceptResult()` doesn't seem like it would be called when there's an exception that leaks out of a command class. Instead, that gets handled in `CommandApp.RunAsync()` where either 1) it gets leaked out to my Program.cs (which is how I currently do it) or 2) gets passed to the provided exception handler, if provided.

Here's the problem I'm dealing with. I use Serilog in my application. I register a factory in DI that I use to create an `ILogger` instance which gets registered as a singleton.

When I handle exceptions in my Program.cs, I do not have access to my underlying `ILifetimeScope`, so I'm forced to use something like `AnsiConsole.WriteException()`. But this would not deliver that exception information to other important sinks I've registered with Serilog (e.g. file logs).

When an exception occurs, I'd like to be able to use an instance of `ILogger` to write the exception information through Serilog, so that it gets delivered to multiple destinations (console included). With the provided exception handler mechanism, this isn't possible since I am not allowed an opportunity to resolve `ILogger` from my DI container.

One way I can think of to allow for this is to have something like `ICommandExceptionHandler` that gets resolved (by Spectre.Console) through DI, giving me an opportunity to resolve dependencies in a constructor that I can use to log exception information. If we can reuse `ICommandInterceptor` for this, that'd be fine too, but given the different levels in the stack trace, I'm not sure if that type is usable at the `CommandApp` level. Design & solution I leave to you, but I can't completely get rid of my exception handling logic from Program.cs (nor my hack) until I'm able to do this.

If you need additional details or review, please don't hesitate to ask!

rcdailey:
One small detail I might add:

If you decide to go with a registered type to handle exception processing, we still need the existing exception handler functionality as a fallback. There's a possibility that, due to DI dependencies and order of execution, that you would fail to resolve instances of `ICommandExceptionHandler` (or whatever you use).

In those cases, the behavior I'd expect is that you silently ignore those failures and treat them as if they are not registered to begin with. You'd proceed to call the configured exception handler (or propagate it, depending on how the user configured things).

This allows me to do `AnsiConsole.WriteException()` (actually, I'd probably just let spectre handle it) in the configured exception handler as a fallback, and use `ILogger` in my exception handler subclass.

So basically I think this means:

1. If registered exception handler objects are available and process an exception, we don't go to the command handler fallback
2. If no exception handlers are registered, OR they fail to be constructed, then we use the exception handler as a fallback.

You might even make #1 a user choice similar to how ASP.NET middleware works (user can say they "handled" the exception or not, which would control whether or not the fallback is used).

All this can probably get pretty complicated so again, I am happy to let you work out the details. I just wanted to brain dump some logistical concerns.

Thanks again for everything.

nils-a:
@rcdailey thank you for the feedback. Regarding exception handling with DI, you might want to have a look at #1411 which should address that.

rcdailey:
Excellent!! With both of these changes, I am able to get rid of all of my hacks. I cherry picked your other PR into this one, did another local build, and tested out the changes on my end.

This is the `Program.cs` I started with (before your two changes):

```cs
internal static class Program
{
    private static ILifetimeScope? _scope;
    private static IBaseCommandSetupTask[] _tasks = Array.Empty<IBaseCommandSetupTask>();
    private static ILogger? _log;

    [SuppressMessage("Design", "CA1031:Do not catch general exception types")]
    public static int Main(string[] args)
    {
        var builder = new ContainerBuilder();
        CompositionRoot.Setup(builder);

        var logLevelSwitch = new LoggingLevelSwitch();
        var appDataPathProvider = new AppDataPathProvider();
        CompositionRoot.RegisterExternal(builder, logLevelSwitch, appDataPathProvider);

        var app = new CommandApp(new AutofacTypeRegistrar(builder, s => _scope = s));
        app.Configure(config =>
        {
        #if DEBUG
            config.PropagateExceptions();
            config.ValidateExamples();
        #endif

            config.Settings.PropagateExceptions = true;
            config.Settings.StrictParsing = true;

            config.SetApplicationName("recyclarr");
            config.SetApplicationVersion(
                $"v{GitVersionInformation.SemVer} ({GitVersionInformation.FullBuildMetaData})");

            var interceptor = new CliInterceptor(logLevelSwitch, appDataPathProvider);
            interceptor.OnIntercepted.Subscribe(_ => OnAppInitialized());
            config.SetInterceptor(interceptor);

            CliSetup.Commands(config);
        });

        var result = 1;
        try
        {
            result = app.Run(args);
        }
        catch (Exception ex)
        {
            if (_log is null)
            {
                AnsiConsole.WriteException(ex, ExceptionFormats.ShortenEverything);
            }
            else
            {
                _log.Error(ex, "Non-recoverable Exception");
            }
        }
        finally
        {
            OnAppCleanup();
        }

        return result;
    }

    private static void OnAppInitialized()
    {
        if (_scope is null)
        {
            throw new InvalidProgramException("Composition root is not initialized");
        }

        _log = _scope.Resolve<ILogger>();
        _log.Debug("Recyclarr Version: {Version}", GitVersionInformation.InformationalVersion);

        _tasks = _scope.Resolve<IOrderedEnumerable<IBaseCommandSetupTask>>().ToArray();
        _tasks.ForEach(x => x.OnStart());
    }

    private static void OnAppCleanup()
    {
        _tasks.Reverse().ForEach(x => x.OnFinish());
    }
}
```

And this is where we are after all the cleanup:

```cs
internal static class Program
{
    [SuppressMessage("Design", "CA1031:Do not catch general exception types")]
    public static int Main(string[] args)
    {
        var builder = new ContainerBuilder();
        CompositionRoot.Setup(builder);

        var app = new CommandApp(new AutofacTypeRegistrar(builder));
        app.Configure(config =>
        {
        #if DEBUG
            config.PropagateExceptions();
            config.ValidateExamples();
        #endif

            // config.Settings.PropagateExceptions = true;
            config.Settings.StrictParsing = true;

            config.SetApplicationName("recyclarr");
            config.SetApplicationVersion(
                $"v{GitVersionInformation.SemVer} ({GitVersionInformation.FullBuildMetaData})");

            config.SetExceptionHandler((ex, resolver) =>
            {
                var log = (ILogger?) resolver?.Resolve(typeof(ILogger));
                if (log is null)
                {
                    AnsiConsole.WriteException(ex, ExceptionFormats.ShortenEverything);
                }
                else
                {
                    log.Error(ex, "Non-recoverable Exception");
                }
            });

            CliSetup.Commands(config);
        });

        return app.Run(args);
    }
}
```

I was able to clean up quite a bit. Also notice I no longer needed the hack in my custom type registrar to access the `ILifetimeScope`. Another nice touch is the `SetExceptionHandler()`, which I can use now, to grab the `ILogger`. It would be nice if there was a version of `Resolve()` that was more statically typed with generics, but not a huge deal.

The other stuff cleaned up nicely as separate interceptor classes:

```cs
public class VersionLogInterceptor(ILogger log) : ICommandInterceptor
{
    public void Intercept(CommandContext context, CommandSettings settings)
    {
        log.Debug("Recyclarr Version: {Version}", GitVersionInformation.InformationalVersion);
    }
}

public class GlobalTaskInterceptor(IOrderedEnumerable<IGlobalSetupTask> tasks) : ICommandInterceptor
{
    public void Intercept(CommandContext context, CommandSettings settings)
    {
        tasks.ForEach(x => x.OnStart());
    }

    public void InterceptResult(CommandContext context, CommandSettings settings, ref int result)
    {
        tasks.Reverse().ForEach(x => x.OnFinish());
    }
}
```

Overall I am pleased with these improvements. Thank you very much for listening to my feedback!

FrankRay78 (approved):

PR #1438, opened 2024-01-30
feat/write list irenderable

### LINKED ISSUE #1437: AnsiConsole.Write receives an IEnumerable from IRenderable
I find it frustrating that I need to do an AnsiConsole.Write when I want to create text that are pieces of library classes.  For example, I want to display "File create, in the path: {PATH} ", where PATH would be an item of type TextPath.

image
My solution would be to allow an overload of AnsiConsole's Write method, which receives a list of IRenderable items and assembles them in one line.

Where the final output would be something like this:
image

Another solution that has been considered would be for items such as Text, TextJson and the like to have a ToString method that returns the string to be used in string interpolations
image

**Additional context**

jos3s:
- new overload of the write method receiving IEnumerable<IRenderable> or IRenderable[] params
- implementing a new overload in the classes that implement IAnsiConsole
- implementing a new overload in the TestConsole
- new unit tests for Write, TextPath, BarChart and Table

fixes #1437

## Changes

I created an overload in the Write method of the IAnsiConsole interface and implemented it in all the classes that implement this interface. The purpose of this overload is to receive a list of IRenderable objects and render them together without having to call AnsiConsole.Write for each IRenderable object.


patriksvensson:
Wouldn't adding an extension method for `IAnsiConsole` that accomplishes the same thing suffice?

```csharp
public static void Write(this IAnsiConsole console, params IRenderable[] renderables)
{
    foreach(var renderable in renderables) 
    {
        console.Write(renderable);
    }
}
```

src/Spectre.Console/IAnsiConsole.cs
```diff
@@ -41,4 +41,16 @@ public interface IAnsiConsole
...
+    /// Writes a <see cref="IRenderable"/> items to the console.
+    /// </summary>
+    /// <param name="renderables">The <see cref="IRenderable"/> items to write.</param>
+    void Write(params IRenderable[] renderables);
+
+    /// <summary>
+    /// Writes a <see cref="IRenderable"/> items to the console.
+    /// </summary>
+    /// <param name="renderables">The <see cref="IRenderable"/> items to write.</param>
+    void Write(IEnumerable<IRenderable> renderables);
```
patriksvensson, line 55: We do not want to extend the `IAnsiConsole` interface with something like this.

jos3s:
Yes, it could be. When analyzing the code, I noticed that AnsiConsole's extension methods always receive pure values ("string", double, int and others), so I thought that in this case, since it will receive an IRenderable, it would be a class method rather than an extension.

I will remove these changes and just add an extension method that receives the IRenderable[] params

jos3s:
@patriksvensson 

I don't know why the build is breaking, I looked for the file displayed in the log and it doesn't exist

> /home/runner/work/spectre.console/spectre.console/test/Spectre.Console.Tests/Unit/Widgets/ProgressBarTests.cs(3,2): error CS0619: 'UsesVerifyAttribute' is obsolete: 'No longer required. Usages of this attribute can be removed.'

FrankRay78:
Looks like a nice clean implementation @jos3s at first glance, I'll prioritise reviewing this in the coming week.

patriksvensson:
@FrankRay78 I've already started the review

Issue #1401, opened 2023-12-05
Error when using C# PublishTrimmed, Trimmer can remove Types and Properties that spectre.console expects the type to have. (


rafaelsc:
**Information**
 - OS: Windows 10
 - Version: 0.48.0
 - Terminal: Any
 - Runtime: NET 8 with Publish Trimmed
 
**Describe the bug**

**To Reproduce**
_I Tried without success creating a standalone minimal console app to reproduce, but look like trimmer is not removing the same types/properties in the same way in my application._

- Create a new Console Application with NET8, with PublishTrimmed, SelfContained, PublishSingleFile, PublishReadyToRun.
- Use `ExceptionFormatter` with a Exception with a stack trace.
- `ExceptionFormatter` can some times throw a `System.NullReferenceException` when using reflection.

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Spectre.Console.ExceptionFormatter.TryGetTupleName(ParameterInfo parameter, Type parameterType, String& tupleName) in /_/src/Spectre.Console/Widgets/Exceptions/ExceptionFormatter.cs:line 314
   at Spectre.Console.ExceptionFormatter.GetParameterName(ParameterInfo parameter) in /_/src/Spectre.Console/Widgets/Exceptions/ExceptionFormatter.cs:line 280
   at Spectre.Console.ExceptionFormatter.GetStackFrames(Exception ex, ExceptionSettings settings) in /_/src/Spectre.Console/Widgets/Exceptions/ExceptionFormatter.cs:line 79
   at Spectre.Console.ExceptionFormatter.GetException(Exception exception, ExceptionSettings settings) in /_/src/Spectre.Console/Widgets/Exceptions/ExceptionFormatter.cs:line 22
   at Program.<>c__DisplayClass0_1.<<Main>$>b__0(LiveDisplayContext ctx)
   at Spectre.Console.LiveDisplay.<>c__DisplayClass15_0.<Start>b__0(LiveDisplayContext ctx) in /_/src/Spectre.Console/Live/LiveDisplay.cs:line 47
   at Spectre.Console.LiveDisplay.<>c__DisplayClass17_0.<<StartAsync>b__0>d.MoveNext() in /_/src/Spectre.Console/Live/LiveDisplay.cs:line 80
--- End of stack trace from previous location ---
[...]

Issue #1155, opened 2023-02-05
Support for .NET NativeAOT


bsdayo:
I'm trying to build a command line tool with `Spectre.Console.Cli` and .NET NativeAOT enabled in .NET 7 (by `<PublishAot>true</PublishAot>` in csproj). But when I publish this project I get the following warnings:
```
MSBuild version 17.4.1+9a89d02ff for .NET
  正在确定要还原的项目…
  所有项目均是最新的，无法还原。
  RenameX -> D:\Code\RenameX\RenameX\bin\Release\net7.0\win-x64\renx.dll
  Generating native code
C:\Users\b1acksoil\.nuget\packages\spectre.console.cli\0.46.1-preview.0.6\lib\net7.0\Spectre.Console.Cli.dll
: warning IL2104: Assembly 'Spectre.Console.Cli' produced trim warnings. For more information see https://aka
.ms/dotnet-illink/libraries [D:\Code\RenameX\RenameX\RenameX.csproj]
C:\Users\b1acksoil\.nuget\packages\spectre.console.cli\0.46.1-preview.0.6\lib\net7.0\Spectre.Console.Cli.dll
: warning IL3053: Assembly 'Spectre.Console.Cli' produced AOT analysis warnings. [D:\Code\RenameX\RenameX\Ren
ameX.csproj]
C:\Users\b1acksoil\.nuget\packages\runtime.win-x64.microsoft.dotnet.ilcompiler\7.0.2\framework\System.Private
.Xml.dll : warning IL3053: Assembly 'System.Private.Xml' produced AOT analysis warnings. [D:\Code\RenameX\Ren
ameX\RenameX.csproj]
C:\Users\b1acksoil\.nuget\packages\runtime.win-x64.microsoft.dotnet.ilcompiler\7.0.2\framework\System.Net.Htt
p.dll : warning IL3053: Assembly 'System.Net.Http' produced AOT analysis warnings. [D:\Code\RenameX\RenameX\R
enameX.csproj]
    正在创建库 bin\Release\net7.0\win-x64\native\renx.lib 和对象 bin\Release\net7.0\win-x64\native\renx.exp
  RenameX -> D:\Code\RenameX\RenameX\bin\Release\net7.0\win-x64\publish\
```

After I run the published exe I get the following error:

```
Error: Could not find a constructor for 'NekoSpace.RenameX.MainCommand'.
```

The minimal reproducable program is like this:

```csharp
using Spectre.Console;
[...]

PR #1508, opened 2024-04-04
WIP Initial Trimming Support

### LINKED ISSUE #955: Multiple errors when trying to ILLink/Trim a console app that uses Spectre.
**Information**
 - OS: Windows
 - Version: 0.44.0
 - Terminal: Windows Terminal

**Describe the bug**
When trying to use Spectre on a console app that targets .NET6, and trying to build a self-contained library, I get multiple ILLink errors, such as IL2070, IL2072, IL2067, IL2072, IL2026, and IL2087.

**To Reproduce**
Reference Spectre.Console 0.44.0 from a console app, use something from Spetre, and build with PublishTrimmed=True, PublishSingleFile=True, "--self-contained -r win-x64 -f net6.0". Multiple errors.

**Expected behavior**
Builds and works fine.

**Additional context**
```
/_/src/Spectre.Console/Cli/Internal/CommandBinder.cs(10,37): Trim analysis error IL2070: Spectre.Console.Cli.CommandBinder.Bind(CommandTree,Type,ITypeResolver): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors' in call to 'System.Type.GetConstructors()'. The parameter 'settingsType' of method 'Spectre.Console.Cli.CommandBinder.Bind(CommandTree,Type,ITypeResolver)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Binding/CommandConstructorBinder.cs(34,9): Trim analysis error IL2072: Spectre.Console.Cli.CommandConstructorBinder.CreateSettings(CommandValueLookup,ConstructorInfo,ITypeResolver): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors' in call to 'System.Activator.CreateInstance(Type,Object[])'. The return value of method 'System.Reflection.MemberInfo.DeclaringType.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Modelling/CommandParameter.cs(76,17): Trim analysis error IL2072: Spectre.Console.Cli.CommandParameter.Assign(CommandSettings,ITypeResolver,Object): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The return value of method 'Spectre.Console.Cli.PairDeconstructorAttribute.Type.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Modelling/CommandParameter.cs(123,17): Trim analysis error IL2072: Spectre.Console.Cli.CommandParameter.Assign(CommandSettings,ITypeResolver,Object): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The return value of method 'Spectre.Console.Cli.CommandParameter.ParameterType.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Binding/CommandPropertyBinder.cs(34,9): Trim analysis error IL2067: Spectre.Console.Cli.CommandPropertyBinder.CreateSettings(ITypeResolver,Type): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The parameter 'settingsType' of method 'Spectre.Console.Cli.CommandPropertyBinder.CreateSettings(ITypeResolver,Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Binding/CommandValueBinder.cs(97,13): Trim analysis error IL2072: Spectre.Console.Cli.CommandValueBinder.GetFlag(CommandParameter,Object): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The return value of method 'Spectre.Console.Cli.CommandParameter.ParameterType.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Binding/CommandValueBinder.cs(48,13): Trim analysis error IL2072: Spectre.Console.Cli.CommandValueBinder.GetLookup(CommandParameter,ITypeResolver,Object): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The return value of method 'Spectre.Console.Cli.PairDeconstructorAttribute.Type.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Binding/CommandValueResolver.cs(133,17): Trim analysis error IL2026: Spectre.Console.Cli.CommandValueResolver.GetConverter(CommandValueLookup,CommandValueBinder,ITypeResolver,CommandParameter): Using member 'System.ComponentModel.TypeDescriptor.GetConverter(Type)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Generic TypeConverters may require the generic types to be annotated. For example, NullableConverter requires the underlying type to be DynamicallyAccessedMembers All.
/_/src/Spectre.Console/Cli/Internal/Binding/CommandValueResolver.cs(133,17): Trim analysis error IL2072: Spectre.Console.Cli.CommandValueResolver.GetConverter(CommandValueLookup,CommandValueBinder,ITypeResolver,CommandParameter): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.All' in call to 'System.ComponentModel.TypeDescriptor.GetConverter(Type)'. The return value of method 'System.Type.GetElementType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Binding/CommandValueResolver.cs(153,17): Trim analysis error IL2072: Spectre.Console.Cli.CommandValueResolver.GetConverter(CommandValueLookup,CommandValueBinder,ITypeResolver,CommandParameter): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.All' in call to 'System.ComponentModel.TypeDescriptor.GetConverter(Type)'. The return value of method 'Spectre.Console.Cli.IFlagValue.Type.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Binding/CommandValueResolver.cs(153,17): Trim analysis error IL2026: Spectre.Console.Cli.CommandValueResolver.GetConverter(CommandValueLookup,CommandValueBinder,ITypeResolver,CommandParameter): Using member 'System.ComponentModel.TypeDescriptor.GetConverter(Type)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Generic TypeConverters may require the generic types to be annotated. For example, NullableConverter requires the underlying type to be DynamicallyAccessedMembers All.
/_/src/Spectre.Console/Cli/Internal/Binding/CommandValueResolver.cs(156,13): Trim analysis error IL2026: Spectre.Console.Cli.CommandValueResolver.GetConverter(CommandValueLookup,CommandValueBinder,ITypeResolver,CommandParameter): Using member 'System.ComponentModel.TypeDescriptor.GetConverter(Type)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Generic TypeConverters may require the generic types to be annotated. For example, NullableConverter requires the underlying type to be DynamicallyAccessedMembers All.
/_/src/Spectre.Console/Cli/Internal/Binding/CommandValueResolver.cs(156,13): Trim analysis error IL2072: Spectre.Console.Cli.CommandValueResolver.GetConverter(CommandValueLookup,CommandValueBinder,ITypeResolver,CommandParameter): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.All' in call to 'System.ComponentModel.TypeDescriptor.GetConverter(Type)'. The return value of method 'Spectre.Console.Cli.CommandParameter.ParameterType.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Binding/CommandValueResolver.cs(34,21): Trim analysis error IL2072: Spectre.Console.Cli.CommandValueResolver.GetParameterValues(CommandTree,ITypeResolver): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The return value of method 'Spectre.Console.Cli.CommandParameter.ParameterType.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Configuration/ConfigurationHelper.cs(21,40): Trim analysis error IL2070: Spectre.Console.Cli.ConfigurationHelper.GetGenericTypeArguments(Type,Type,Type[]&): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.Interfaces' in call to 'System.Type.GetInterfaces()'. The parameter 'type' of method 'Spectre.Console.Cli.ConfigurationHelper.GetGenericTypeArguments(Type,Type,Type[]&)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/DefaultPairDeconstructor.cs(63,9): Trim analysis error IL2067: Spectre.Console.Cli.DefaultPairDeconstructor.GetConverter(Type): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.All' in call to 'System.ComponentModel.TypeDescriptor.GetConverter(Type)'. The parameter 'type' of method 'Spectre.Console.Cli.DefaultPairDeconstructor.GetConverter(Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/DefaultPairDeconstructor.cs(63,9): Trim analysis error IL2026: Spectre.Console.Cli.DefaultPairDeconstructor.GetConverter(Type): Using member 'System.ComponentModel.TypeDescriptor.GetConverter(Type)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Generic TypeConverters may require the generic types to be annotated. For example, NullableConverter requires the underlying type to be DynamicallyAccessedMembers All.
/_/src/Spectre.Console/Cli/Internal/DefaultPairDeconstructor.cs(33,17): Trim analysis error IL2067: Spectre.Console.Cli.DefaultPairDeconstructor.Spectre.Console.Cli.IPairDeconstructor.Deconstruct(ITypeResolver,Type,Type,String): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The parameter 'valueType' of method 'Spectre.Console.Cli.DefaultPairDeconstructor.Spectre.Console.Cli.IPairDeconstructor.Deconstruct(ITypeResolver,Type,Type,String)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Widgets/Exceptions/ExceptionFormatter.cs(408,77): Trim analysis error IL2070: Spectre.Console.ExceptionFormatter.<TryResolveStateMachineMethod>g__GetDeclaredMethods|13_0(IReflect): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods', 'DynamicallyAccessedMemberTypes.NonPublicMethods' in call to 'System.Reflection.IReflect.GetMethods(BindingFlags)'. The parameter 'type' of method 'Spectre.Console.ExceptionFormatter.<TryResolveStateMachineMethod>g__GetDeclaredMethods|13_0(IReflect)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Cli/Internal/Composition/Activators.cs(111,37): Trim analysis error IL2070: Spectre.Console.Cli.ReflectionActivator.GetGreediestConstructor(Type): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors' in call to 'System.Type.GetConstructors()'. The parameter 'type' of method 'Spectre.Console.Cli.ReflectionActivator.GetGreediestConstructor(Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Internal/TypeConverterHelper.cs(26,9): Trim analysis error IL2087: Spectre.Console.TypeConverterHelper.GetTypeConverter<T>(): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.All' in call to 'System.ComponentModel.TypeDescriptor.GetConverter(Type)'. The generic parameter 'T' of 'Spectre.Console.TypeConverterHelper.GetTypeConverter<T>()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
/_/src/Spectre.Console/Internal/TypeConverterHelper.cs(26,9): Trim analysis error IL2026: Spectre.Console.TypeConverterHelper.GetTypeConverter<T>(): Using member 'System.ComponentModel.TypeDescriptor.GetConverter(Type)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Generic TypeConverters may require the generic types to be annotated. For example, NullableConverter requires the underlying type to be DynamicallyAccessedMembers All.
/_/src/Spectre.Console/Cli/Internal/TypeResolverAdapter.cs(28,13): Trim analysis error IL2067: Spectre.Console.Cli.TypeResolverAdapter.Resolve(Type): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The parameter 'type' of method 'Spectre.Console.Cli.TypeResolverAdapter.Resolve(Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
```

phil-scott-78:
* Marked Spectre.Console and Spectre.Console.Cli as trimmable to enable warnings
* Marked Demo.cspoj and Injection.cspoj to be PublishedAOT and to include all trim warnings

***

### Spectre.Console
* Marked Spectre.Console methods that used TypeConverter with suppressed messages
* Marked ExceptionFormatter as requiring dynamic code and added a fallback for AOT scenarios.

***
### Spectre.Console.Cli
* Settings were discovered automatically with reflection. Added a new method for adding commands that allows explicit configuration of the settings to prevent them from being trimmed. Marked the old method as requiring dynamic code and pointed users to the new overload
* Added attributes to the default TypeRegister to mark the types as being instantiated dynamically.
* Suppressed a whole slew of warnings with command bulding and running the commands. With the commands and their settings being marked as having DynamicallyAccessedMembers this, theoretically, should work without error.

Biggest worry is in CommandParameterComparer. The MetadataToken is stripped when published as AOT so it can't be used. A simple equal seems to work the same, but I'm not sure the historical reason MetadataToken is used here. @patriksvensson, you remember?

I only have a handful of projects to test with myself. I worked through the Injection.csproj and Demo.cspoj in the CLI examples and ensured they worked properly. Also verified `cli explain` and `cli xmldoc` still worked when trimming.

This would close #955, #1401 and #1155. Those issues have a lot of good info to get me going, and any feedback on this from @azchohfi,  @Gnbrkm41, @CyberSinh , @ricardoboss and @Simonl9l 

Also, I apologize for all the `#if NET6_0_OR_GREATER`. Working around netstandard support really uglied some of these method signatures up.


phil-scott-78:
Disregard the failed unit tests, that was due to multiple AddCommands and I was lost in the weeds thinking I broke something with that property comparer. Those are fixed, but I'm still convinced I broke something with the comparer, but I'm not sure what...

examples/Cli/Demo/Commands/Run/RunCommand.cs
```diff
@@ -49,7 +50,7 @@ public sealed class Settings : CommandSettings
         [CommandOption("--verbosity <VERBOSITY>")]
         [Description("Set the MSBuild verbosity level. Allowed values are q[grey]uiet[/], m[grey]inimal[/], n[grey]ormal[/], d[grey]etailed[/], and diag[grey]nostic[/].")]
         [TypeConverter(typeof(VerbosityConverter))]
-        [DefaultValue(Verbosity.Normal)]
+        [DefaultValue("normal")]
```
phil-scott-78, line 53: This seems to have just been a bug in the demo. Not sure it has worked in a while. `Verbosity.Normal` was passed into the converter as an integer of `2` which the type converter kicked out. Tried in both AOT and regular scenarios.

patriksvensson: I think we should update the `VerbosityConverter` to convert the integer to the enum.

patriksvensson:
Awesome work. Now comes the big question, how do we know what to update when we make changes and/or additions. Otherwise the trimming support will slowly become more and more unusable. Could we document this somewhere?

examples/Cli/Demo/Commands/Run/RunCommand.cs
phil-scott-78: Actually, I think I found a bug in the .NET runtime related to AOT.  Open up my IDE for the first time in half a year and I'm already blowing stuff up

https://github.com/dotnet/runtime/issues/100688

phil-scott-78:
MS confirmed the bug with NativeAOT. I cleaned up the commit to be explicit about why there is a work around, separate from the main AOT work.

Regarding the future of maintenance once this is enabled, with `<IsTrimmable>true</IsTrimmable>` set, the compiler will alert us to what we need to do moving forward.  If you add code that uses reflection, it'll warn you to make sure you mark your code appropriately with the `DynamicallyAccessedMembers` attribute. Really the only way things could get screwy is if someone went into a method marked to suppress warnings and started using reflection on types we hadn't considered yet. A scenario like that is `FlagValue` where it is never instantiated directly so its constructor was getting trimmed out. 

The biggest maintenance headache is trying to read the code, especially the generic argument definitions, with the `#IF` statements to work around NetStandard support.

But give me a bit to grab a beer and I'll add additional comments throughout, especially around FlagValue and where we've suppressed messages.

patriksvensson:
@phil-scott-78 Sounds good to me!

agocke:
My recommendation for making the cleaner is using the polysharp nuget package and switch on the flag to generate runtime attributes. 

I’ve also been working on annotation so I’ll take a look at this later. I’ll say that I’m pretty skeptical that CLI can be made safe without a source generator.

phil-scott-78:
Love the idea of polysharp.

src/Spectre.Console/Internal/TypeConverterHelper.cs
```diff
@@ -49,6 +51,10 @@ public static bool TryConvertFromStringWithCulture<T>(string input, CultureInfo?
         }
     }
 
+#if NET6_0_OR_GREATER
+    [UnconditionalSuppressMessage("AssemblyLoadTrimming", "IL2026", Justification = TypeConverterWarningsCanBeIgnored)]
```
agocke, line 55: I don't think this will work. The problem is that `GetConverter` is a really complex API under the covers. Aside from looking for attributes recursively, it also looks for ICustomTypeDescriptor, which does a huge amount of reflection.

Is there a way we can narrow the acceptable input types here?

src/Spectre.Console.Cli/Internal/Binding/CommandConstructorBinder.cs
```diff
@@ -2,6 +2,10 @@ namespace Spectre.Console.Cli;
 
 internal static class CommandConstructorBinder
 {
+#if NET6_0_OR_GREATER
+    [UnconditionalSuppressMessage("AssemblyLoadTrimming", "IL2072",
+        Justification = TrimWarnings.SuppressMessage)]
+#endif
```
agocke, line 8: Why is this suppressed? Shouldn't it be RUC? Same for a lot of the other suppressions.

src/Spectre.Console/Internal/TypeConverterHelper.cs
agocke: @patriksvensson I think this is a question for your API design and policy on breaking changes. We've spent a lot of time looking over TypeConverter for the past few months since we've been working on trimming WinForms and there doesn't seem to be a way to make `ICustomTypeDescriptor` trim-safe.

I think we _might_ be able to get a method that does basically the same stuff without `ICustomTypeDescriptor` support, but that involves a new API in ComponentModel and a TFM bump at the minimum (so, at least .NET 9).

Alternatively, if you only care about supporting a few types in this API, or only a few ways of getting type converters via reflection, I think we could change this API to throw if there's something unexpected. That would meet the general guidance of trimming, which is that your API shouldn't behave differently before and after trimming without a warning being present.

There's an advanced feature that we call "feature switches" that could enable the behavior switch only if the user has expressed "trimming intent" by putting `PublishTrimmed` or `PublishAot` in their project file. That would be an option if you're concerned about taking a breaking change, but it is kind of unfortunate because someone who adds trimming later on would suddenly get the new behavior without warning.

agocke (commented):
Example of using poly sharp

phil-scott-78:
I'm sold on polysharp. Added it and reworked the PR to use it.

patriksvensson:
If we're going to use PolySharp, we should probably remove the other polyfill libraries that we use.

patriksvensson:
Also, might be a good idea to create a new demo where trimming is enabled and leave the current demo code as-is.

phil-scott-78:
The final four games were pretty boring last night, and the boys went to bed in a reasonable manner, so I had time to implement @agocke's suggestion on polysharp and tweaking some of the attributes. Pushed up those changes.

I'll create a new DemoAOT project. I think that's a good idea. Not only does it keep the original there, having one explicitly with AOT will lead to (hopefully) better discovery.

phil-scott-78:
Added the new DemoAOT project to demonstrate how to configure the project. One hang up right now is the string constructor trick @0xced included a while back where we automatically convert types with a single param string constructor (e.g. `DirectoryInfo`). I added an example on how to work around that, but it'd be nice if there was a source generator that could produce those required attributes to keep the compiler from stripping those dynamically called constructors out.

patriksvensson:
@phil-scott-78 Is this ready for review, or do you have more things to add/fix/change?

agocke:
One more thing to consider is, rather than a demo app, you can use a "test app" to have the linker/aot compiler verify all the dependencies. Instructions are at https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0#show-all-warnings-with-test-app but the key part is adding `TrimmerRootAssembly` that causes the trimmer to treat all entry points into a library as being reachable.

Here's the code for an AOT test app that I wrote and put in the `test` folder:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>

    <PublishAot>true</PublishAot>
    <!-- Show warnings from dependencies, not just from the app -->
    <TrimmerSingleWarn>false</TrimmerSingleWarn>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="../../src/Spectre.Console.Cli/Spectre.Console.Cli.csproj" />
    <TrimmerRootAssembly Include="Spectre.Console" />
    <TrimmerRootAssembly Include="Spectre.Console.Cli" />
  </ItemGroup>

</Project>
```

phil-scott-78:
Before review, I definitely want to try out that @agocke test project. I had a few minutes to use a computer today and added it only to find a few more warnings.

agocke:
I’ve tried to do some annotation on my own. I could share it but don’t know the best way. Comments in this PR? New PR?

agocke:
Pushed all my changes to https://github.com/agocke/spectre.console/tree/cli-trimmable. There are a lot of CLI annotations, but I'm not sure there's a simple alternative.

codymullins:
@agocke pulled in your code to my project - it looks like [we need to add `[DynamicallyAccessedMembers]`](https://github.com/codymullins/spectre.console/commit/3ae5dfb66dca96ff7a4a9e7e8d3677531ac633ff) to the `ITypeRegistrar`. I didn't see this attribute being used already, are we doing this in some other way?

Without the above, trimming doesn't seem to work when integrating with `Microsoft.Extensions.DependencyInjection`:

```csharp
public class DependencyInjectionRegistrar(IServiceCollection services) : ITypeRegistrar, IDisposable
{
    // ... omitted code

    public void Register(Type service, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type implementation)
    {
        Services.AddSingleton(service, implementation);
    }

    // ...
}
```

agocke:
@codymullins  did you also pull the project file changes? Polysharp adds the polyfill for the attributes.

codymullins:
I forked the repo...not familiar with Polysharp. Let me investigate if I forked anything up 😄

agocke:
Wanted to give an update on where I am with this: I think I've got the annotations in a good place locally, and they make plain Spectre.Console usable, minus the changes to TypeConverter.

Spectre.Console.Cli is another story. The annotations make it unusable -- all the current APIs are not trim safe. Unfortunately, I don't see a way to use the existing APIs to manually configure it, either. Reflection is deeply baked into the API surface.

What I've started is defining a new parallel set of APIs that allow a combination of configuration and reflection, that could eventually be implemented by a source generator.

Right now I'm kind of stuck on Bind. There's both a lot of reflection in there and a lot of currently internal surface area, so lifting out a new API is taking a while.

An intermediate state that I think would be good is just annotating the current surface. That would make it clear what's usable and what's not. And we need to define a path forward for the TypeConverter behavior in Spectre.Console. The existing TypeConverter is a non-starter.

phil-scott-78:
I happened to have a chance to look at this today and I also landed at your idea of the intermediate step. I don't think AOT perfection is possible but given a handful of well-placed annotations it does become quite difficult to break things. Trickiest thing, to me at least, is the [magic](https://github.com/spectreconsole/spectre.console/blob/main/src/Spectre.Console.Cli/Internal/Binding/CommandValueResolver.cs#L219) for trying to call a constructor for a type with a single string param constructor. 

Other than that, most cases seem to chug along, right? Intrinsic types have their TypeConverter handled pretty well in .NET 8. Custom types we've always pushed people to annotate their fields with a `TypeConverter` annotation which has the right annotations itself for including the proper `DynamicallyAccessedMemberTypes`. Once we get past that, we've never really supported anything overly complex out of the box like `List<T>` or anything beyond arrays of intrinsic types, which seem to work fine. 

Of course, there might be countless other things I'm missing here. Could you provide some examples of things that absolutely would not work with the Cli project because of the current `TypeConverter`? Even just documenting those might be a good step, and maybe improve the error messages too.

agocke:
The problem is that debugging AOT-specific problems is basically impossible. Our guidance for AOT-safety is to either guarantee behavior, or marking it as unsafe. Suppressing warnings that then bite users during deployment isn't really doing any favors.

I do think I can build a new API for the CLI layer, it's just going to take a while.

agocke:
> Could you provide some examples of things that absolutely would not work with the Cli project because of the current TypeConverter?

So, the big problem with TypeConverter is just that `GetConverter` is fundamentally unsafe due to reliance on `ICustomTypeDescriptor` as a fallback. My alternate implementation works around this by implementing a `GetConverter` method that only uses the `TypeConverter` attribute and doesn't respect `ICustomTypeDescriptor`.

Beyond that we still run into problems in binding. GetConverter is a good example

```C#
if (parameter.ParameterType.IsArray)
{
    // Return a converter for each array item (not the whole array)
    var elementType = parameter.ParameterType.GetElementType();
    if (elementType == null)
    {
        throw new InvalidOperationException("Could not get element type");
    }

    return (TypeDescriptor.GetConverter(elementType), GetStringConstructor(elementType));
}
```

Let's say you replace that `GetConverter` call with a call to my "safe" version that has `DynamicallyAccessedMembers` on it. That doesn't help because the annotations don't allow you to annotate array element types, just the array itself. So the return type of `parameter.ParameterType.GetElementType()` will never satisfy those annotations. What we can do instead is pass along an extra piece of information: as part of the parameter we not only keep the parameter Type, we also keep the element Type. The constructor for `CommandParameter` would require you to pass both.

The trick for this is it just becomes very verbose for the user, and really removes the benefit of doing reflection at all. The user ends up typing a bunch of boilerplate code. Hence the source generator.

FrankRay78:
Setting aside 20 years of .Net coding and still only being able to partially follow the above discussion....

I'm missing the use case for AOT in this scenario @agocke (@phil-scott-78) - who would want to trim `Spectre.Console` and for what purpose, and on what platforms?

Performance tweaks for a CLI application don't make sense, so I'm left with native compiled .Net apps on Linux/Mac, avoiding the need to install (then maintain) the .Net SDK comes to mind. 

Am broadly correct in my guess work?

A wider question is the suitability of AOT for prime time?  The above discussion seems fraught with code massaging unfamiliar to bog standard .Net development practices. 

Interested in where this might go.

slang25:
I use Spectre.Console in my Native AOT CLI I distribute at work. It's actually a perfect use-case for AOT, you typically want AOT for apps that need to start fast, a non-AOT CLI feels sluggish.

AnonymousRetard:
> I use Spectre.Console in my Native AOT CLI I distribute at work. It's actually a perfect use-case for AOT, you typically want AOT for apps that need to start fast, a non-AOT CLI feels sluggish.

Same here. Startup time without AOT can easily be more than a second and with AOT it can reach a 1/10th of that, makes the CLI feel a lot more responsive. I also tested making another CLI with another purpose using System.CommandLine which does support trimming fully and then I was able to make a complete self-contained AOT binary in just a few MB while my Spectre.Console AOT CLI is more like ~25MB if I remember correctly. It would be very nice if I could do the same which Spectre.Console as I liked it a lot more apart from not having full trim support. A smaller binary should reduce startup times even more because then there's much less data that needs to be read from disk when first starting up.

FrankRay78:
Thank you @slang25, @AnonymousRetard for your comments - I'm obviously not well-informed about AOT, so it's good to hear actual user feedback.

phil-scott-78:
I'll add my two-sense, I got interested in this when I decided to see if I could make an oh-my-posh like prompt in .Net. I don't have my results handy, but my "time to first rendering of some yellow text and a smiley face" (TTFROSYTASF) benchmark started at like 140ms and dropped to something below 40ms. That's a huge difference for that usage scenario.

Now, the question as to whether it is worth the developer headache: My personal opinion right now would be to put the effort in the main Spectre.Console library to make it compatible, but then mark the Spectre.Console.Cli project as explicitly not supported due to reflection with a `RequiresDynamicCode` attribute on the `app.Run` methods and be done with it. I'm not sure that project can ever support it without some very involved refactoring

phil-scott-78:
dropping this effort for a new one...

FrankRay78:
A good effort @phil-scott-78, none the less.

PR #1637, opened 2024-09-11
Improve exception if a (multi)selection prompt is used incorrectly

0xced:
Before this commit, the selection prompt would throw an `InvalidOperationException` (Sequence contains no elements) and the multi selection prompt would throw an `ArgumentOutOfRangeException` (Index was out of range. Must be non-negative and less than the size of the collection.)

Both would occur because the prompts were never meant to be empty.

Maybe we should change the `SelectionPrompt` and `MultiSelectionPrompt` constructors to always require at least one item so that this situation could never occur? But that would be a breaking change and should be discussed in a separate issue. This one is not breaking and just provides a more understandable error message.

---


patriksvensson (approved):
This makes perfect sense. LGTM 👍

Issue #1472, opened 2024-02-19
ElapsedTimeRemaining Keeps updating 


mazorg:
**Information**
 - OS: Windows 11/Ubuntu 22.04
 - Version: 0.48
 - Terminal: Windows terminal, Powershell, ubuntu terminal

**Describe the bug**
The ElapsedTimeRemaining Colums wont stop updating for stopped tasks.

**To Reproduce**
Change Line 24 of Program.cs in Progress example from `new RemainingTimeColumn()` to `new ElapsedTimeColumn()`

**Expected behavior**
The elapsed time should stop progressing as the progresstask finishesh.

PR #1672, opened 2024-10-24
This fixes issue #1472 to stop ElapsedTimeRemaining from updating

jbennink:
This fixes issue #1472. I read the issue and thought it should not be too hard and perfrect for a first PR. It turns out it is caused due to StopTime not being set when the maxValue is reached. I simply added one line to fix this which fixed the issue.

- [X] No documentation changes are required.

## Changes

The issue occurs due to StopTime not being set when the MaxValue is readed. The Isfinished bool checks both StopTime and Value > MaxValue, but since the ElapsedTime simple checks whether StopTime is set, and that is never set except when StopTask is called this keeps on counting. By adding the assignment to StopTime in the Update method when _value > _maxValue this fixes the issue.

P.S. I did not comment on the issue since it is not a behaviour change but simply a bugfix.

---

### DIFF
--- a/src/Spectre.Console/Live/Progress/ProgressTask.cs
+++ b/src/Spectre.Console/Live/Progress/ProgressTask.cs
@@ -201,6 +201,7 @@ public sealed class ProgressTask : IProgress<double>
             if (_value > _maxValue)
             {
                 _value = _maxValue;
+                StopTime ??= DateTime.Now;
             }
 
             var timestamp = DateTime.Now;


FrankRay78:
Hello @jbennink, thanks for taking the time to submit this, and write up the PR description. I have only taken a cursory look, and whilst it's a one-line change (not to diminish the significance of such things), we need all changes to have unit test coverage. Have you ensured there is before/after coverage of the change, and/or should we still be expecting tests to follow for this PR?

PR #1675, opened 2024-11-03
Adding Enricher for Azure Pipelines

### LINKED ISSUE #1666: Markup Not Working in Azure Pipelines
**Information**
 - OS: Linux Ubuntu 22.04.5 build agent
 - Version: Spectre.Console 0.49.1
 - Terminal: Azure pipeline output

**Describe the bug**
When a console app built using Spectre.Console is run from an Azure pipeline PowerShell task, the output does not honour the colours set.

**To Reproduce**
1. Create a simple .NET console application and reference Spectre.Console
2. Execute this single line in the console application:
```csharp
AnsiConsole.MarkupLine("[green]This should be green but isn't[/]");
```
3. Run the console application from a `PowerShell@2`:
```yaml
- task: PowerShell@2
  displayName: 'Test Script'
  inputs:
    targetType: 'inline'
    script: |
      Write-Host "`e[32mThis should be green and is`e[0m"
      test-console-application
```

**Expected behavior**
Both lines written to the console should be green. The first is using an ANSI escape code, and the second is using Spectre.Console. However, as seen in the screenshot, only the first line is green.

**Screenshots**
image

---

BlazeFace:
fixes #1666 

## Changes

Added an enricher with same settings as GitHub Actions, looks for TF_BUILD, used [Variables](https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml) as reference. This will be set to true in build tasks.

---

### DIFF
--- a/src/Spectre.Console/Enrichment/CI/AzurePipelinesEnricher.cs
+++ b/src/Spectre.Console/Enrichment/CI/AzurePipelinesEnricher.cs
@@ -1,0 +1,24 @@
+namespace Spectre.Console.Enrichment;
+
+internal sealed class AzurePipelinesEnricher : IProfileEnricher
+{
+    public string Name => "AzurePipeline";
+
+    public bool Enabled(IDictionary<string, string> environmentVariables)
+    {
+        if (environmentVariables.TryGetValue("TF_BUILD", out var value))
+        {
+            return value?.Equals("true", StringComparison.OrdinalIgnoreCase) ?? false;
+        }
+
+        return false;
+    }
+
+    public void Enrich(Profile profile)
+    {
+        profile.Capabilities.Ansi = true;
+        profile.Capabilities.Legacy = false;
+        profile.Capabilities.Interactive = false;
+        profile.Capabilities.Links = false;
+    }
+}
--- a/src/Spectre.Console/Enrichment/CI/AzurePipelinesEnricher.cs
+++ b/src/Spectre.Console/Enrichment/CI/AzurePipelinesEnricher.cs
@@ -6,12 +6,8 @@ internal sealed class AzurePipelinesEnricher : IProfileEnricher
 
     public bool Enabled(IDictionary<string, string> environmentVariables)
     {
-        if (environmentVariables.TryGetValue("TF_BUILD", out var value))
-        {
-            return value?.Equals("true", StringComparison.OrdinalIgnoreCase) ?? false;
-        }
-
-        return false;
+        environmentVariables.TryGetValue("TF_BUILD", out var environmentValue);
+        return string.IsNullOrWhiteSpace(environmentValue);
     }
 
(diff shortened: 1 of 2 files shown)


src/Spectre.Console/Enrichment/CI/AzurePipelinesEnricher.cs
```diff
@@ -0,0 +1,24 @@
...
+
+internal sealed class AzurePipelinesEnricher : IProfileEnricher
+{
+    public string Name => "AzurePipeline";
+
+    public bool Enabled(IDictionary<string, string> environmentVariables)
+    {
+        if (environmentVariables.TryGetValue("TF_BUILD", out var value))
+        {
+            return value?.Equals("true", StringComparison.OrdinalIgnoreCase) ?? false;
```
patriksvensson, line 11: No need to check if it's `true` or not here. A simple `!string.IsNullOrWhitespace` should suffice.

patriksvensson (changes requested):

src/Spectre.Console/Enrichment/CI/AzurePipelinesEnricher.cs
BlazeFace: Updated!

BlazeFace: Just for clarity we don't need to use the if statement on the TryGet since in the case of the variable not being present environmentValue will be the default value for a string, so the null the check will return true, thus the method will return false.

src/Spectre.Console/Enrichment/CI/AzurePipelinesEnricher.cs
```diff
@@ -6,12 +6,8 @@ internal sealed class AzurePipelinesEnricher : IProfileEnricher
...
     public bool Enabled(IDictionary<string, string> environmentVariables)
     {
-        if (environmentVariables.TryGetValue("TF_BUILD", out var value))
-        {
-            return value?.Equals("true", StringComparison.OrdinalIgnoreCase) ?? false;
-        }
-
-        return false;
+        environmentVariables.TryGetValue("TF_BUILD", out var environmentValue);
+        return string.IsNullOrWhiteSpace(environmentValue);
```
patriksvensson: This expression needs to be inverted, otherwise the enricher will always be enabled.

patriksvensson (changes requested):

src/Spectre.Console/Enrichment/CI/AzurePipelinesEnricher.cs
BlazeFace: Should be updated with the inversion!

patriksvensson (approved):

patriksvensson:

PR #1696, opened 2024-11-23
Removed the Nullable package reference in favor of PolySharp

phil-scott-78:
Removed the Nullable package references as its functionality was duplicated in PolySharp. Also upgraded PolySharp to the latest while I was in there.

I gave it a go seeing if `TunnelVisionLabs.ReferenceAssemblyAnnotator` happened to be made redundant, but the IL weaving it does into the old assemblies is still being relied upon.


patriksvensson (approved):

patriksvensson:
Merged 👍

PR #1765, opened 2025-02-20
Fix IndexOutOfRangeException in ExceptionFormatter

kvpt:
Commit b2689be introduced a regression when stacktrace don't have frame.

Here is the crash in my app.
image

The change that caused this issue is that stackTrace.GetFrame(0) has been replaced by stackTrace.GetFrames()[0].
Method GetFrame does index bound check whereas stackTrace.GetFrames()[0] is a simple array access and don't.
I put back the previous method to fix the issue.

I tried to make a test to cover this case but I failed to create an exception with no frame.


kvpt:
Fixed by PR #1800.

PR #1789, opened 2025-03-31
Added startup interceptors

### LINKED ISSUE #1788: Startup interceptors

I want to be able to print something every time my cli starts. With Spectre.Console.Cli, I would need to do that in every command handler instead of one, single place.

I gave it a thought and think the ability to register a `IStartupInterceptor` in the type registrar would be nice. It would then be called directly after the command was processed but before the command is executed.

As I said earlier, without this, I would need to add this print (or anything else I want for every command) to each command individually. This is unnecessary, error prone and noisy.

**Additional context**

The `IStartupInterceptor` is akin to the `ICommandInterceptor`, so the concept itself is not foreign to the project.

In addition, it would be nice if both, the `IStartupInterceptor` and the `ICommandInterceptor` would be able to access the `ITypeResolver`, so they can run access any services they need without constructor injection.

---

ricardoboss:
fixes #1788

## Changes

Added a new interface `IStartupInterceptor` that is called before any command is executed. It can be registered using the type registrar and receives a context object, which also has access to the type registrar.

I also added a property to the `ICommandContext` so that it can access the type registrar.

---

I know not all checkboxes above are checked and I know I didn't open a discussion first. I plan to still add documentation and tests. Since I'll open this as a draft, you can already give me feedback on the design before I do all the accompanying work.

---

### DIFF
--- a/src/Spectre.Console.Cli/Internal/CommandExecutor.cs
+++ b/src/Spectre.Console.Cli/Internal/CommandExecutor.cs
@@ -59,6 +59,10 @@ internal sealed class CommandExecutor
         // Create the resolver.
         using (var resolver = new TypeResolverAdapter(_registrar.Build()))
         {
+            var startupInterceptors = resolver.Resolve(typeof(IEnumerable<IStartupInterceptor>)) as IEnumerable<IStartupInterceptor>;
+            var startupContext = new StartupContext { TypeResolver = resolver };
+            startupInterceptors?.ForEach(i => i.Intercept(startupContext));
+
             // Get the registered help provider, falling back to the default provider
             // if no custom implementations have been registered.
             var helpProviders = resolver.Resolve(typeof(IEnumerable<IHelpProvider>)) as IEnumerable<IHelpProvider>;
@@ -94,7 +98,8 @@ internal sealed class CommandExecutor
                 arguments,
                 parsedResult.Remaining,
                 leaf.Command.Name,
-                leaf.Command.Data);
+                leaf.Command.Data,
+                resolver);
 
             // Execute the command tree.
             return await Execute(leaf, parsedResult.Tree, context, resolver, configuration).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


FrankRay78:
Cool. I suggest adding some unit tests to demonstrate this functionality @ricardoboss.

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -59,6 +59,10 @@ public async Task<int> Execute(IConfiguration configuration, IEnumerable<string>
         // Create the resolver.
         using (var resolver = new TypeResolverAdapter(_registrar.Build()))
         {
+            var startupInterceptors = resolver.Resolve(typeof(IEnumerable<IStartupInterceptor>)) as IEnumerable<IStartupInterceptor>;
+            var startupContext = new StartupContext { TypeResolver = resolver };
```
patriksvensson, line 63: No need to create a StartUpContext if there are no startup interceptors. Add a `.ToArray` after the resolver and check if `Length > 0` and do the rest

src/Spectre.Console.Cli/Internal/CommandExecutor.cs
```diff
@@ -59,6 +59,10 @@ public async Task<int> Execute(IConfiguration configuration, IEnumerable<string>
...
+            startupInterceptors?.ForEach(i => i.Intercept(startupContext));
```
patriksvensson, line 64: Use a normal `foreach` loop here

src/Spectre.Console.Cli/StartupContext.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Spectre.Console.Cli;
+
+public sealed class StartupContext
+{
+    public required ITypeResolver TypeResolver { get; init; }
```
patriksvensson, line 5: Use an explicit constructor which accept a `ITypeResolver` and checks it for null.

patriksvensson (changes requested):
All warnings need to be fixed as well.

ricardoboss:
@FrankRay78 @patriksvensson I made the changes you asked for. This still needs to be documented and I think maybe an easier way to register it would be nice. What do you think?

PR #1804, opened 2025-04-11
Fix for live display renderer recording.

### LINKED ISSUE #1723: Exporting recorded output of a Live display can give empty results.
Hello everyone 👋. I was exploring this awesome library and started to improve my personal CLI tools with it, but unfortunately I have found an issue, which I wanted to share.

**Information**
 - OS: Windows
 - Version: 0.49.1
 - Terminal: Windows Terminal

**Describe the bug**
I have observed that live display of a table is not always properly recorded. For relatively small tables, when I call `AnsiConsole.ExportText()` after recording, it returns an empty string. However for large tables, it returns string with the full table output (as expected).

**To Reproduce**
Example C# program to simulate issue:

```
using Spectre.Console;

AnsiConsole.Record();

var table = new Table().Centered();

AnsiConsole
    .Live(table)
    .AutoClear(false)
    .Start(ctx =>
    {
        table.AddColumn("Foo");
        table.AddColumn("Bar");

        for (int i = 0; i < 5; i++)
        {
            table.AddRow($"Foo {i}", $"Bar {i}");
        }
    });


var text = AnsiConsole.ExportText();
Console.WriteLine(text);
```
**Expected behavior**
In the example the value of a `text` is empty, but should contain an exported table output (the expected result of the whole example program is to print the same table twice). When I increase number of rows to for ex. 50 or 100, it produces the expected result.

**Additional context**
I want to be able to save my program live outputs into a file. Piping live outputs from shell is not possible/supported, and it throws `System.IO.IOException` (as described in this issue comment: https://github.com/spectreconsole/spectre.console/issues/1393#issuecomment-1841178111), so the only reasonable way to do it seems to be: record, export and save into a file.

**Possible fix/workaround**
I have looked into the source code and found that method `LiveDisplayRenderer.Completed(bool autoclear)` forces to redraw when overflow occurs, which causes, the output to be correctly recorded and exported for larger tables. 
As a workaround solution I have created small fix that turn this redraw behavior on, every time when recorder has been registered. You can see it on my fork: https://github.com/spectreconsole/spectre.console/compare/main...bmic86:spectre.console:fix/recording_on_live

---

bmic86:
fixes #1723

## Changes

Fixed issue with exporting live display outputs as described in the issue #1723

---

### DIFF
--- a/src/Spectre.Console/Live/LiveDisplayRenderer.cs
+++ b/src/Spectre.Console/Live/LiveDisplayRenderer.cs
@@ -26,7 +26,7 @@ internal sealed class LiveDisplayRenderer : IRenderHook
             }
             else
             {
-                if (_context.Live.HasRenderable && _context.Live.DidOverflow)
+                if (_context.Live.HasRenderable && (_context.Live.DidOverflow || _context.Live.IsRecording))
                 {
                     // Redraw the whole live renderable
                     _console.Write(_context.Live.RestoreCursor());
--- a/src/Spectre.Console/Live/LiveRenderable.cs
+++ b/src/Spectre.Console/Live/LiveRenderable.cs
@@ -16,6 +16,7 @@ internal sealed class LiveRenderable : Renderable
     public bool HasRenderable => _renderable != null;
     public VerticalOverflow Overflow { get; set; }
     public VerticalOverflowCropping OverflowCropping { get; set; }
+    public bool IsRecording => _console is Recorder;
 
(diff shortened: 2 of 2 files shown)


bmic86:


sob., 12 kwi 2025 o 01:30 microsoft-github-policy-service[bot] <
***@***.***> napisał(a):

patriksvensson:
I'm sorry, but I can't merge this.
We really don't want special logic around recording in the widgets.

PR #1813, opened 2025-04-27
Preserve links when markup is being truncated/linebroken

### LINKED ISSUE #363: Links (id and target) do not survive the render process
**Information**
 - Version: most recent master: 6a5c507936f602764464d2c358c8cf17da36374b

**Describe the bug**
If a link is output into a small grid for example it gets layouted into multiple lines. The way the linkId is currently implemented (Hash+Random) will cause those segments to be treated as different links. This results in a situation where the hover effect occurs only on parts of the link.

**To Reproduce**
```csharp
// create a narrow grid with width 10 containing a link
AnsiConsole.Console.Write(
    new Grid() { Width = 10 }
        .AddColumn()
        .AddRow("[link]https://google.com[/]")
);

// now hovering over the link (in a terminal that supports hyperlinks) only highlights part of the link.
// also because of the implicit link markup the link target is wrong as well.
```

**Expected behavior**
Even if the link is split into multiple rows during the render process it should be the same link (same id so the terminal recognizes it as the same link) and the target should still be valid for all segments.

**Screenshots**
image

**Additional context**
I think this are really two bugs.
1) The link target is wrong because the implicit link is not set inside the style but pretty late in `AnsiBuilder.GetAnsi` but at that point the inner text might already be split and layouted all over the place.
2) The link Id should probably be persisted inside Style to survive the layout process. At the moment everything is split into segments and later a linkId is assigned.

**Ideas:**
Regarding point 1) if a link markup only has text but not set an explicit link the assignment should happen before the style is created (probably in `StyleParser.cs`)
Regarding point 2) IMO the linkId should be part of the Style and just be incremented and set in the constructor if the provided link was not null or empty. The counter could simply be a static field and used like that: `LinkId = Interlocked.Increment(ref _linkId);`

To be honest I don't get why the current LinkId hasher was implemented as it is. First a hash is generated based on the link target and than a random number is added, so why hash the link in the first place? Just incrementing a counter would be simpler and deterministic and would result in a more testable output.

ShaunLawrie:
fixes #363

- [ ] All existing tests are still running without errors

## Changes

This doesn't fix the ID issues talked about in the original issue, only the link target being incorrect when the segment is split over multiple lines or truncated and the link is to be generated via the text e.g. `[link]https://example.com[/]`.

Attempting to share a link ID over these portions makes things much more complex and the end result is just that the link will have a correct hover effect on some terminals. This could be addressed in a separate change.

---


patriksvensson:
Hi! 

Thanks so much for taking the time to submit this PR. 
I really appreciate the contribution!

Unfortunately, I haven’t had the capacity to review this in a timely manner, and since it was opened, the repository has changed significantly. Because of that, this PR is no longer in a state where it can be reviewed or merged as-is. This is not your fault, but mine, and I'm sorry about this.

I’m going to close it for now to help keep things tidy. That said, if you’re still interested in contributing, please feel free to open a new pull request that’s up to date with the current state of the repository, and I’ll be happy to take a look when I can. *I'll promise I'll do better this time around.*

Thanks again for your work and for your interest in the project!

PR #1835, opened 2025-06-10
Resolved CLI: Add negation flags like clang or gcc

### LINKED ISSUE #1701: CLI: Add negation flags like clang or gcc
Passing a false value use a negation like `--no-execution` instead of `--execution false`

Mo-Sanchez:
Fixes #1701 

## Changes

This pull request implements support for **negation flags** using the `--no-` prefix, similar to the behavior found in compilers like Clang and GCC.

- Added support to set the flag value to `"false"` when prefixed with `--no-`
- Adjusted `CommandTreeParser.cs` logic to strip the `no-` prefix before looking up matching options
 - If a boolean flag such as `--verbose` is defined, the parser will now also accept `--no-verbose` to automatically assign the value `false`
- Added a unit test in `CommandAppTests.FlagValues.cs` to verify the behavior with a `BoolSettings` class.

### DIFF
--- a/src/Spectre.Console.Cli/Internal/Parsing/CommandTreeParser.cs
+++ b/src/Spectre.Console.Cli/Internal/Parsing/CommandTreeParser.cs
@@ -256,7 +256,13 @@ internal class CommandTreeParser
         if (context.State == State.Normal)
         {
             // Find the option.
-            var option = node.FindOption(token.Value, isLongOption, CaseSensitivity);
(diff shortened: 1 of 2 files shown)


patriksvensson:
Before we can review this PR you need to do two things:

* Add the check list you've removed to the PR description and make sure that all criterias are fullfilled
* Agree to the CLA

Feel free to re-open this PR once these two things have been done.

PR #1836, opened 2025-06-10
Resolved: Esc Key to exit selection

### LINKED ISSUE #851: Esc Key to exit selection

Mo-Sanchez:
Fixes #851 

Changes
+ Added logic in Selection Prompt to detect when the Escape key is pressed and return the default value, effectively exiting the prompt early.
+ Included unit test by simulating an Escape key input and asserting the result is null


patriksvensson:
Before we can review this PR you need to do two things:

* Add the check list you've removed to the PR description and make sure that all criterias are fullfilled
* Agree to the CLA

Feel free to re-open this PR once these two things have been done.

PR #1942, opened 2025-10-31
Fixes issue with Panel not applying overflow to children

### LINKED ISSUE #1940: Panel not respecting IOverflowable
Given this

```csharp
var markup = new Markup("[yellow]foo[/] [red]pneumonoultramicroscopicsilicovolcanoconiosis[/] [blue]bar qux[/]")
    .Overflow(Overflow.Ellipsis);
var panel = new Panel(markup)
{
    Width = 20,
};

// When
AnsiConsole.Write(panel);
```

I'm seeing

```
┌───────────────────────────────────────────────────────────┐
│ foo pneumonoultramicroscopicsilicovolcanoconiosis bar qux │
└───────────────────────────────────────────────────────────┘
```

Expected
```
┌──────────────────┐
│ foo              │
│ pneumonoultrami… │
│ bar qux          │
└──────────────────┘
```

I have a fix, just want to make sure this is an actual bug or designed behavior with panels

phil-scott-78:
Open to other suggestions on how to resolve, but feels like this is right. While I was in there I added some other tests around Grid and Table with IOverflowable because I wasn't sure about their behavior either.

Resolves #1940


patriksvensson (approved):
LGTM 👍

PR #1995, opened 2026-01-05
Add source generation for colors, emojis and spinners

phil-scott-78:
This was suspiciously easy. Included in the project are the old generated emojis/colors/spinners and a unit test verifying anything the original is in the source generated. This should be dropped eventually, but figured nice to have while reviewing the code. 

Beyond ensuring all the values are the same, I did try my best to make this as performant as I know how. Reading a mb of emojis in json is obviously something we don't want to be doing all the time, but I have it flowing through pipelines that are performant so it should hopefully avoid that. On my machine, with Rider and VS Code I saw it generate the code about 20 times over about 5 minutes as I was editing and even when it ran it was only a 50ms hit. 

This also drops the Generator projects, the Asciicast stuff is no longer needed either with the new docs.


patriksvensson (approved):
Looks good to me! 👍

PR #2016, opened 2026-02-17
Text prompt editable default

### LINKED ISSUE #595: Text prompt with editable default value
Console input is a chore, it'd be nice to be able to pre-fill the user's answer with a default value so that the user can just hit ENTER to proceed to the next question. 

I would love the ability to have a prompt that takes a default value, and that makes that editable to the user, which works on PC, Mac, and Linux. 

---

MapelSiroup:
fixes #595

- [x] I have commented on the issue above and discussed the intended changes [ my proposal for #595 ](https://github.com/spectreconsole/spectre.console/issues/595#issuecomment-3917086132)
- [x] A maintainer has signed off on the changes and the issue was assigned to me

## Changes

This PR allows the insertion of the default value inside text field/input buffer making it editable and already available instead of the current DefaultValue behavior to simply set the text variable with the default when empty.

1. `TextPrompt.cs`: Added a `bool DefaultInput { get; set; }` property and, when showing the prompt, if enabled the default value is placed into the editable input field.

2. `TextPromptExtensions.cs`: Added matching extension `DefaultInput<T>(this TextPrompt<T> obj, bool state)` to enable/disable the editable default behavior when creating TextPrompts.

3. `AnsiConsoleExtensions.Input.cs`: Added a new ReadLine() overload that accepts an `initialInput (string?)` and enqueues a ConsoleKeyInfo per character. The overload consumes those queued key events first, then falls back to `console.Input.ReadKeyAsync()`.

4. `TextPromptTests.cs`: Added a unit test `Should_Return_Error_If_Default_Choice_Invalid` that checks the default input system (Testing the validity checks whether the input matches a valid choice in the list).

## Affected files:
- src/Spectre.Console/Prompts/TextPrompt.cs
- src/Spectre.Console/Prompts/TextPromptExtensions.cs
- src/Spectre.Console/Extensions/AnsiConsoleExtensions.Input.cs
- src/Spectre.Console.Tests/Unit/Prompts/TextPromptTests.cs
- src/Spectre.Console.Tests/Expectations/Prompts/Text/InvalidDefaultChoice.Output.verified.txt

---


MapelSiroup:

src/Spectre.Console/Extensions/AnsiConsoleExtensions.Input.cs
```diff
@@ -5,6 +5,100 @@ namespace Spectre.Console;
 /// </summary>
 public static partial class AnsiConsoleExtensions
 {
+    internal static async Task<string> ReadLine(this IAnsiConsole console, Style? style, bool secret, char? mask, IEnumerable<string>? items = null, CancellationToken cancellationToken = default, string? initialInput = null)
```
patriksvensson: I don't think we should duplicate the code here; we should use one `ReadLine`.

patriksvensson (changes requested):

patriksvensson:
@MapelSiroup Seems to be a merge conflict now that needs to be resolved

src/Spectre.Console/Prompts/TextPrompt.cs
```diff
@@ -58,6 +58,11 @@ public sealed class TextPrompt<T> : IPrompt<T>, IHasCulture
     /// </summary>
     public bool ShowDefaultValue { get; set; } = true;
 
+    /// <summary>
+    /// Gets or sets the default value input method. Defaults to true which places the default value in the input buffer.
+    /// </summary>
+    public bool DefaultInput { get; set; }
```
patriksvensson: I'm not sure why this is needed. The naming does not tell very well what I can expect to happen.

Couldn't this be removed, and we could use `ShowDefaultValue` instead?

patriksvensson (changes requested):

patriksvensson (approved):

PR #2028, opened 2026-02-25
Feature/cancel prompt

### LINKED ISSUE #851: Esc Key to exit selection

### LINKED ISSUE #1392: Allow user to cancel input during Prompt()
When requesting user input multiple times, it would be nice to provide a way for the user to return to previous steps in case they've entered some wrong information.

We could capture the ESC keypress during `AnsiConsole.Prompt` and set a variable, e.g. `AnsiConsole.Prompt(prompt, out var cancelled)` to indicate the user pressed escape. Or, the return value of  `AnsiConsole.Prompt` could be `null` if the input was cancelled.

There are some incomplete code samples [here](https://stackoverflow.com/questions/31996519/listen-on-esc-while-reading-console-line) that I've tried, but there are multiple issues with the code such as exceptions and formatting problems. It would be great if this was a native solution in Spectre.Console.

red-stronghold:
Fixes #851 
Fixes #1392 

## Changes

For `SelectionPrompt<T>`:
- Allow setting a `Func<T>` property to be triggered on Cancel
- Extension methods to set the property, either with a Func or with a Value (for easier use)
- Capture 'ESC' key and cancel the prompt
- Return the specified result if the prompt is cancelled
- Unit tests to cover both extension methods.

For `MultiSelectionPrompt<T>`:
- Allow setting a `Func<List<T>>` property to be triggered on Cancel
- Extension methods to set the property, either with a Func, a List, a Value or an empty List (for easier use)
- Capture 'ESC' key and cancel the prompt
- Return the specified result if the prompt is cancelled
- Unit tests to cover all extension methods.

If property not set, the prompts works as before.

This is a re-make of the closed PR #1919 
The old PR was closed due to time.
[reason](https://github.com/spectreconsole/spectre.console/pull/1919/#issuecomment-3936197210)

---

### DIFF
--- a/src/Spectre.Console/Prompts/List/ListPrompt.cs
+++ b/src/Spectre.Console/Prompts/List/ListPrompt.cs
@@ -67,7 +67,7 @@ internal sealed class ListPrompt<T>
                 {
                     break;
                 }
-                else if(result == ListPromptInputResult.Abort)
+                else if (result == ListPromptInputResult.Abort)
                 {
                     state.Cancel();
                     break;


patriksvensson:
@red-stronghold I rebased your PR for you. Will review soon.

PR #2091, opened 2026-04-16
Fix link parsing to terminate properly

### LINKED ISSUE #2090: Links do not terminate properly and multiple links per markup block don't work
**Information**
 - OS: Windows 11 25H2
 - Version: 0.55.0
 - Terminal: Windows Terminal

**Describe the bug**
Links seems to be fundamentally broken currently as they are not properly terminated in AnsiMarkup.Parse. The parser does not consider the active link while producing AnsiMarkupSegments. It lacks any behavior to terminate a link once started and multiple links in the same blob are not supported.

**To Reproduce**
Basic link usage in markup produces incorrect results:
```cs
ansiConsole.Markup("[blue link=https://some.address/]Some text[/] and [link=https://another.address/]some more text[/].\n");
```
This results in the entire line being a single link to some.address because the first link never terminates. Only the first link is ever started and then is copied into all subsequent AnsiMarkupSegments no matter what in that Markup() call.

**Expected behavior**
Links should be correctly terminated and links inside links should even work.

**Screenshots**
```cs
        program.ansiConsole.Markup("[blue link=https://google.com/]Some text[/][link=https://bing.com/]some more text[/].\n[link=https://bing.com/]some more text[/].\n");
```
This test code produces only a single multi-line link:

Image

---

zhuman:
Fixes #2090

- [X] I have checked that there isn't already another pull request that solves the above issue
- [X] All newly added code is adequately covered by tests

I used GitHub Copilot to generate the new tests.

## Changes

This change is pretty straightforward - it promotes links to be treated similarly to styles in the markup parser, where they are parsed out of new tags and pushed into a stack. Closing the tag caused them to get popped from the stack. The AnsiMarkupSegments comparison that is used to merge together text of the same style now compares the active link.

Three new tests were added to SelectionPrompt confirming that links embedded in SelectionPrompts don't exhibit the incorrect behavior (this is where I originally noticed the issues). Two new tests were added to AnsiConsoleTests.Markup to validate that nesting and link closure works.

---

### DIFF
--- a/src/Spectre.Console.Ansi/AnsiMarkup.cs
+++ b/src/Spectre.Console.Ansi/AnsiMarkup.cs
@@ -54,7 +54,8 @@ public sealed class AnsiMarkup
         using var tokenizer = new MarkupTokenizer(markup);
 
         var result = new List<AnsiMarkupSegment>();
-        var stack = new Stack<Style>();
+        var styleStack = new Stack<Style>();
+        var linkStack = new Stack<Link?>();
         var link = default(Link?);
 
         while (tokenizer.MoveNext())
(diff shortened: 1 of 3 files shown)


patriksvensson (approved):

patriksvensson:

PR #2100, opened 2026-04-24
Option to exclude vertical padding for live progress renderer

### LINKED ISSUE #1726: Make it possible to exclude linebreaks before and after Progress bars.
When using `Progress`, Spectre enforces a linebreak before and after the progress bar. This gets quite annoying if we use 2 separate progress bars after each other since that will include 2 empty linebreaks between the progress bars.

I do not know if there is a technical requirement for forcing these linebreaks. But i would much prefer to just add a linebreak if neccessary myself.

Some kind option for disabling forced linebreaks around progress bars.

If we need two separate progress bars after eachother, i have resolved to using `AnsiConsole.Cursor.Move(CursorDirection.Up, 1)` after the first progress bar in order to remove the duplicate linebreak.

**Additional context**
image

---

james-newell-forge:
Fixes: https://github.com/spectreconsole/spectre.console/issues/1726

No AI was used in commit

## Changes

Currently, a live progress widget forces you to have a 1-top, 1-bottom vertical padding and there is no option to change this.  This commit adds a new option to `Progress.cs` called `ExcludeVerticalPadding` that when set excludes the Padder added in `DefaultProgressRenderer.cs` (default false to keep existing behavior).

### DIFF
--- a/src/Spectre.Console/Live/Progress/Progress.cs
+++ b/src/Spectre.Console/Live/Progress/Progress.cs
@@ -268,17 +268,29 @@ public static class ProgressExtensions
     }
 
     /// <summary>
+    /// Sets whether or not to exclude the vertical padding in the display.
+    /// If enabled, the progress display will not include vertical padding.
+    /// </summary>
+    /// <param name="progress">The <see cref="Progress"/> instance.</param>
+    /// <param name="enabled">Whether or not to exclude the vertical padding.</param>
+    /// <returns>The same instance so that multiple calls can be chained.</returns>
+    public static Progress ExcludeVerticalPadding(this Progress progress, bool enabled)
+    {
+        ArgumentNullException.ThrowIfNull(progress);
+
+        progress.ExcludeVerticalPadding = enabled;
+
+        return progress;
+    }
+
+    /// <summary>
     /// Excludes the vertical padding in the display.
     /// </summary>
     /// <param name="progress">The <see cref="Progress"/> instance.</param>
     /// <returns>The same instance so that multiple calls can be chained.</returns>
     public static Progress ExcludeVerticalPadding(this Progress progress)
     {
-        ArgumentNullException.ThrowIfNull(progress);
-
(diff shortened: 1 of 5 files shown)


james-newell-forge:
@microsoft-github-policy-service agree company="ForgeGlobal"

src/Spectre.Console/Live/Progress/Progress.cs
```diff
@@ -259,4 +266,20 @@ public static Progress HideCompleted(this Progress progress, bool enabled)
...
     }
+
+    /// <summary>
+    /// Sets whether or not to exclude the vertical padding in the display.
+    /// If enabled, the progress display will not include vertical padding.
+    /// </summary>
+    /// <param name="progress">The <see cref="Progress"/> instance.</param>
+    /// <param name="enabled">Whether or not to exclude the vertical padding.</param>
+    /// <returns>The same instance so that multiple calls can be chained.</returns>
+    public static Progress ExcludeVerticalPadding(this Progress progress, bool enabled)
```
patriksvensson: We should add two new methods that call this one to make the API consistent:

* `ExcludeVerticalPadding(this Progress progress)`
* `IncludeVerticalPadding(this Progress progress)`

patriksvensson (changes requested):

src/Spectre.Console/Live/Progress/Progress.cs
patriksvensson: Great, but the original method with bool enabled should still be there. The other two should call that one.

james-newell-forge: ah ok, I should have read your comment more carefully.  Thanks for clarifying.

james-newell-forge: Added back `ExcludeVerticalPadding(bool)` and changed the overloaded methods to call it

patriksvensson (approved):

PR #2101, opened 2026-04-25
Fix Align measure to respect explicitly set width

### LINKED ISSUE #1910: Align Measure always return maxWidth even when I explicitly set Width
**Information**
 - OS: Windows
 - Terminal: [e.g Windows Terminal]

**Describe the bug**
Align doesn't override measure from `Renderable`, so it's measured width is always maxWidth. It bothers me when using Align inside Panel with border (Panel border will be expanded no matter if I set Panel width or not)

**Potential solution**
If it's not by-design, maybe considering override Measure?

```
    protected override Measurement Measure(RenderOptions options, int maxWidth)
    {
        var width = Math.Min(Width ?? maxWidth, maxWidth);
        var height = Height ?? options.Height;
        var measurement = _renderable.Measure(options with { Height = height }, width);
        return new Measurement(measurement.Min, width);
    }
```
---

GrantTotinov:
Fixes #1910

## Changes

`Align` did not override the `Measure` method, causing it to always return `maxWidth` regardless of any explicitly set `Width`.

Added a `Measure` override that respects the explicitly set width, consistent with how `Render` already handles it.

Added regression test to ensure explicit width is enforced.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,7 @@ src/.idea/**/uiDesigner.xml
 [Dd]ebug/
 [Rr]elease/
 x64/
+*.lscache
 *_i.c
 *_p.c
 *.ilk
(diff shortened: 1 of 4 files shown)


GrantTotinov:

patriksvensson:
Added a commit to this PR that makes the build pass.

patriksvensson (approved):

patriksvensson:

PR #2114, opened 2026-05-04
Added Markup clearing options and Cursor movement

### LINKED ISSUE #1570: More Clearing Methods
Currently, no, but I realized this could become an issue while planning future projects.

**What I need**: More clearing options, such as clearing the entire line, the whole screen, the end of the screen, and the beginning of the screen.

There are three possible ways to implement this. I prefer the third one as it seems cleaner and more implementable.

1. **Enum-based solution**: Add a specific enum that would be passed when using `AnsiConsole.Clear(enum)`. It could be named `ClearingRange` and could have values like `EntireScreen`, `CurrentLine`, `ToEndOfLine`.
2. **Static class solution**: Replace the `Clear` method in `AnsiConsole` with a `Clear` static class that has static methods like `EntireScreen`, `CurrentLine`, `ToEndOfLine`. However, this syntax doesn't look nice and is too long.
3. **Markup key solution**: Add a specific markup key like `[clear line]` or `[clear:line]`, `[clear:to end of line]` (or possibly `to eol`).

Alternatively, a simpler solution would be to implement a method like `AnsiSequences()` in `AnsiConsole` that returns a single read-only instance for a specific console with as many ANSI codes as possible, possibly grouped into subclasses or more methods. It could work like this:
```
AnsiConsole.AnsiSequences().Clear().Line
// would return `ESC[2K` only if the current terminal supports it, otherwise an empty string
```

I think the markup solution is the best and clearest. However, it might not make sense to use `[clear:line] Some very short line` as it could be confusing.

Using ANSI codes as they are, but this might break functionality when not supported. Check `AnsiConsole.Profile.Capabilities` every time or just once and then, if true, printing the code is too messy and long. It works, but there are much cleaner solutions.

**Additional context**
This feature might not seem useful, but it would be especially helpful when rendering more complex lines that need to be updated often, without flickering.

---

MapelSiroup:
This Pull Request Adds new markup token to manipulated the console cursor, as well as clear parts of the console, such as lines, entire screen or to and from areas. These markup tokens are all being self-closing so no need to encapsulate it.

Fixes #1570 

AI Agent used:
- Copilot Raptor mini 1x : in-line code completion

## Changes

Added These new self-closing markup tokens:
1. [escnext]         Escape the next self-closing token (does not write any output to the console)
2. [wnext]            Escapes and Writes the next self-closing token as raw text
3. [move:to #]     Move cursor to position
4. [clear:to #]     Clear line/area of console 

[Move:] and [Clear:] support these formats:
'to (tag)' such as end of screen (eos), end of line (eol), numeric line number
'x;y', 'home', 'origin' (valid only for move for now) (home and origin are aliases for coordinate 1;1)
Relative tags: 'right', 'forward', 'backward', 'left', 'up', 'down' (valid only for move for now) (home and origin are aliases for coordinate 1;1)
'screen' (clear only, clears the entire screen)

---

### DIFF
--- a/src/Spectre.Console.Ansi.Tests/AnsiMarkupTests.cs
+++ b/src/Spectre.Console.Ansi.Tests/AnsiMarkupTests.cs
@@ -254,6 +254,108 @@ public sealed class AnsiMarkupTests
                 .ShouldBe(expected);
         }
 
+        [Theory]
+        [InlineData("[clear:to eol]", "\x1B[K")]
+        [InlineData("[clear:eol]", "\x1B[K")]
+        [InlineData("[clear:screen]", "\x1B[H\x1B[2J")]
+        [InlineData("[clear:to eos]", "\x1B[0J")]
+        [InlineData("[clear:#23]", "\x1B[23;1H")]
+        public void Should_Output_Expected_Ansi_For_Clear_Directives(string text, string expected)
+        {
+            // Given
+            var fixture = new AnsiFixture();
+
+            // When
+            fixture.Markup.Write(text);
+
+            // Then
+            fixture.Output
+                .ShouldBe(expected);
+        }
+
+        [Theory]
+        [InlineData("[move:to 10]", "\x1B[10;1H")]
+        [InlineData("[move:10;20]", "\x1B[10;20H")]
+        [InlineData("[move:up 3]", "\x1B[3A")]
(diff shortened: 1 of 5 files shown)


patriksvensson:
Thanks for the effort you put into this, but I don't want us to add anything not related to pure styling and links to the markup parser. I'm sorry, but I'm going to close this PR. Thanks for understanding.

PR #2116, opened 2026-05-04
allow validation chaining

### LINKED ISSUE #1742: TextPrompt.Validate does not chain
**Information**
 - OS: Windows
 - Version: 0.49.1
 - Terminal: Windows Terminal

**Describe the bug**
TextPrompt.Validate() does not chain with itself. Only the last call to Validate() will run. Any other validators are ignored. This is clear in TextPromptExtensions where obj.Validator is simply replaced each time Validate() is called.

**To Reproduce**
Chain multiple Validate() calls together. Only the last call will run, regardless of any previous calls.

**Expected behavior**
If Validate() is used multiple times, each Validator should run. Otherwise, comments should be revised to make it clear that only 1 call to Validate() is allowed.

**Additional context**
XML comments in TextPromptExtensions indicate that Validate() should chain, and a comment in TextPrompt says "Run all validators". To me, this indicates that multiple validators are allowed and should work as expected.

Multiple Validate() calls would be useful to separate validation logic, and especially to tailor validation errors more specifically. With a single call, multiple checks all have to run at once and the error message must suggest all possible errors.

Semantics are debatable: should validation stop on the first failure, or run all validators and present all failures? Could be configurable.

I am willing to implement this and open a PR if this is a desirable change.

---

AntekOlszewski:
Fixes #1742 

- [X] I have checked that there isn't already another pull request that solves the above issue
- [X] All newly added code is adequately covered by tests

## Changes

This change was previously proposed in https://github.com/spectreconsole/spectre.console/pull/1973 but became outdated.
This PR reintroduces the same fix on top of the current codebase.
Changes from outdated PR - tests splited into separate ones, since 1 test checked 3 different things, logic unchanged.

---

### DIFF
--- a/src/Spectre.Console.Tests/Unit/Prompts/TextPromptTests.cs
+++ b/src/Spectre.Console.Tests/Unit/Prompts/TextPromptTests.cs
@@ -677,4 +677,68 @@ public sealed class TextPromptTests
         // Then
         result.ShouldBeEquivalentTo(ValidationResult.Success());
     }
+
+    [Fact]
+    public void Validate_MixedOverloads_WithThreeValidators_Returns_ThirdValidationError()
+    {
+        // Given
+        var prompt = new TextPrompt<string>("Enter:");
+        var secondInvoked = false;
+        var thirdInvoked = false;
+
+        prompt
+            .Validate(s => s.Length >= 3, "too short")
+            .Validate(s =>
+            {
+                secondInvoked = true;
+                return s.Contains("a")
+                    ? ValidationResult.Success()
+                    : ValidationResult.Error("missing a");
+            })
+            .Validate(s =>
+            {
+                thirdInvoked = true;
+                return s.EndsWith("z");
+            }, "must end with z");
+
+        // When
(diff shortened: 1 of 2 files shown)


src/Spectre.Console.Tests/Unit/Prompts/TextPromptTests.cs
```diff
@@ -489,4 +489,192 @@ public async Task Uses_case_insensitive_comparison_when_no_comparer_is_passed(st
...
+
+    [Fact]
+    public void Validate_MixedOverloads_ShortCircuits()
+    {
+        // Given
+        var prompt = new TextPrompt<string>("Enter:");
+        var secondInvoked = false;
+
+        prompt
+            .Validate(s => s.Length >= 3, "too short")            .Validate(s =>
```
patriksvensson, line 627:
```suggestion
            .Validate(s => s.Length >= 3, "too short").Validate(s =>
```

patriksvensson (changes requested):
Can we add tests for three chained validations as well?

patriksvensson (approved):

PR #2134, opened 2026-06-05
Fix indeterminate progress completion

### LINKED ISSUE #1907: Indeterminate ProgressTasks should ignore Value/MaxValue
**Information**
Platform independent

**Describe the bug**
ProgressTasks that are set to be indeterminate will show as completed if Value ever reaches MaxValue. This includes the task being hidden automatically if HideCompleted is set to true.

**To Reproduce**
Take enumerating a folder for files, for example. I don't know how many files exist until the process is complete. It's indeterminate. In a producer/consumer pattern, I might be processing the files as they are discovered. If the producer is faster than the consumer, there's no problem. MaxValue will increase faster than Value. However, if the two tasks are fairly balanced or if the consumer is faster, Value will match MaxValue.

The end result of this is that the indeterminate task will often show as completed. With HideCompleted the task will flicker as it is constantly being shown and hidden.

**Expected behavior**
I would think that completion calculations: percentage, the progress bar showing as full and colored as green, the task being hidden if HideCompleted is set, etc. should ignore Value/MaxValue if the task is set to be indeterminate. x

---

marko1olo:
## Summary

Fix indeterminate progress tasks so they do not become completed merely because `Value >= MaxValue`.

For indeterminate tasks, completion now depends on `StopTask()`, which prevents `HideCompleted` flicker and prevents the progress bar from rendering as a completed bar while the task is still active.

## Test plan

- `git show --check --oneline HEAD`
- `git diff --check HEAD~1 HEAD`

I could not run the focused `dotnet test` locally because the repo pins .NET SDK `10.0.300`, while this machine currently has `10.0.202`.

Fixes #1907.

### DIFF
--- a/src/Spectre.Console.Tests/Unit/Live/Progress/ProgressTaskTests.cs
+++ b/src/Spectre.Console.Tests/Unit/Live/Progress/ProgressTaskTests.cs
@@ -106,4 +106,36 @@ public sealed class ProgressTaskTests
         task.Speed.ShouldNotBeNull();
         task.Speed!.Value.ShouldBe(10.0, tolerance: 0.001);
     }
+
+    [Fact]
+    public void IsFinished_Should_Ignore_Value_When_Task_Is_Indeterminate()
+    {
+        // Given
+        var task = new ProgressTask(1, "Foo", 100)
+        {
+            IsIndeterminate = true,
+            Value = 100,
+        };
+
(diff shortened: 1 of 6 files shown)


patriksvensson:
@marko1olo CLA needs to be signed. Feel free to reopen once this has been done.

PR #2137, opened 2026-06-06
Fix table width collapse hang

### LINKED ISSUE #2131: Infinite loop in `TableMeasurer.CollapseWidths` when wrappable columns have width 0
## Information
 - OS: Windows 11
 - Version: 0.55.2
 - Terminal: Windows Terminal

## Description

The `TableMeasurer.CollapseWidths` method enters an **infinite loop** when:
1. One or more wrappable columns already have width `0` (e.g., via `GetColumnWidth()` returning `0`)
2. One or more non-wrappable columns (`NoWrap = true`) have content that exceeds the available console width

This causes the progress display to freeze indefinitely, blocking the refresh thread.

## Steps to Reproduce

```csharp
using Spectre.Console;
using Spectre.Console.Rendering;

// Create a custom column that returns width 0 when hidden
public class ConditionalProgressColumn(ProgressColumn inner, Func<bool> shouldRender) : ProgressColumn
{
    public override IRenderable Render(RenderOptions options, ProgressTask task, TimeSpan deltaTime)
        => shouldRender() ? inner.Render(options, task, deltaTime) : Text.Empty;

    public override int? GetColumnWidth(RenderOptions options)
        => shouldRender() ? inner.GetColumnWidth(options) : 0;  // Returns 0 when hidden!
}

// Use it with a very long task description
await AnsiConsole.Progress()
    .AutoClear(true)
    .Columns(
        new SpinnerColumn(),
        new ConditionalProgressColumn(new ProgressBarColumn(), () => false),  // Hidden, width = 0
        new ConditionalProgressColumn(new PercentageColumn(), () => false),   // Hidden, width = 0
        new TaskDescriptionColumn())  // NoWrap = true
    .StartAsync(async ctx =>
    {
        var task = ctx.AddTask("Short description");

        task.Description = "Short description";
        await Task.Delay(1000);
        
        // Trigger the bug with a very long description
        task.Description = new string('A', 280);  // Very long text
        
        await Task.Delay(5000);
    });
```

## Expected Behavior

The progress display should render correctly, with the long description being truncated or wrapped according to the available space.

## Actual Behavior

The application **freezes** completely. The progress refresh thread enters an infinite loop in `TableMeasurer.CollapseWidths()` and never updates the display again.

## Root Cause

**File:** `Spectre.Console/Widgets/Table/TableMeasurer.cs`  
**Method:** `CollapseWidths` (lines 173-204)

The algorithm tries to reduce wrappable columns to fit the available width. When wrappable columns are already at width `0`, the following happens:

```csharp
// Line 182-187
var maxColumn = widths.Zip(wrappable, (first, second) => (width: first, allowWrap: second))
    .Where(x => x.allowWrap)
    .Max(x => x.width);
// maxColumn = 0 (wrappable columns are already at 0)

var secondMaxColumn = widths.Zip(wrappable, (width, allowWrap) => 
    allowWrap && width != maxColumn ? width : 1).Max();
// secondMaxColumn = 1 (fallback value)

var columnDifference = maxColumn - secondMaxColumn;
// columnDifference = 0 - 1 = -1  ← NEGATIVE!

// Line 190
if (!ratios.Any(x => x != 0) || columnDifference == 0)  // ← BUG: Should be <= 0
{
    break;
}
```

The break condition checks for `columnDifference == 0`, but when `maxColumn = 0`, this becomes `-1`, which doesn't trigger the break. The loop continues indefinitely because `Ratio.Reduce` with negative `maxReduce` values doesn't change the widths.

## Proposed Fix

**Change line 190 in `TableMeasurer.cs`:**

```diff
- if (!ratios.Any(x => x != 0) || columnDifference == 0)
+ if (!ratios.Any(x => x != 0) || columnDifference <= 0)
```

This correctly handles the case where `columnDifference` becomes negative, preventing the infinite loop.

## Workaround

Until this is fixed, avoid returning `0` from `GetColumnWidth()`. Return `1` or `null` instead:

```csharp
public override int? GetColumnWidth(RenderOptions options)
    => shouldRender() ? inner.GetColumnWidth(options) : 1;  // Use 1 instead of 0
```

---

marko1olo:
## Summary

- Stop `TableMeasurer.CollapseWidths` when the largest wrappable column can no longer be reduced.
- This prevents the zero-width wrappable-column case from looping indefinitely when non-wrappable content still exceeds the available table width.
- Add a regression test that renders a narrow table with an already-collapsed wrappable column and long non-wrappable content.

Fixes #2131.

## Tests

- `C:\hades\.dotnet-10.0.300\dotnet.exe test src\Spectre.Console.Tests\Spectre.Console.Tests.csproj --framework net10.0 --filter "FullyQualifiedName~Should_Not_Hang_When_Wrappable_Column_Is_Already_Collapsed" --no-restore`
- `C:\hades\.dotnet-10.0.300\dotnet.exe test src\Spectre.Console.Tests\Spectre.Console.Tests.csproj --framework net10.0 --filter "FullyQualifiedName~Spectre.Console.Tests.Unit.TableTests" --no-restore`
- `git diff --check`

AI-assisted change, reviewed locally before submission.

### DIFF
--- a/src/Spectre.Console.Tests/Unit/Widgets/Table/TableTests.cs
+++ b/src/Spectre.Console.Tests/Unit/Widgets/Table/TableTests.cs
@@ -625,6 +625,27 @@ public sealed class TableTests
     }
 
     [Fact]
+    [GitHubIssue("https://github.com/spectreconsole/spectre.console/issues/2131")]
+    public void Should_Not_Hang_When_Wrappable_Column_Is_Already_Collapsed()
+    {
+        // Given
+        var console = new TestConsole().Width(20);
+        var table = new Table()
+            .HideHeaders()
+            .HideFooters();
+
+        table.AddColumn(new TableColumn("Hidden").Width(0));
+        table.AddColumn(new TableColumn("Description").NoWrap());
+        table.AddRow(string.Empty, new string('A', 80));
+
+        // When
+        console.Write(table);
+
+        // Then
+        console.Output.ShouldNotBeEmpty();
+    }
+
+    [Fact]
     [Expectation("Render_Multiline")]
     public Task Should_Render_Table_With_Multiple_Rows_In_Cell_Correctly()
     {
--- a/src/Spectre.Console/Widgets/Table/TableMeasurer.cs
+++ b/src/Spectre.Console/Widgets/Table/TableMeasurer.cs
@@ -187,7 +187,7 @@ internal sealed class TableMeasurer : TableAccessor
                 var columnDifference = maxColumn - secondMaxColumn;
 
                 var ratios = widths.Zip(wrappable, (width, allowWrap) => width == maxColumn && allowWrap ? 1 : 0).ToList();
-                if (!ratios.Any(x => x != 0) || columnDifference == 0)
+                if (!ratios.Any(x => x != 0) || columnDifference <= 0)
                 {
                     break;
                 }


patriksvensson:
@marko1olo You will need to sign the CLA for this to be reviewed

patriksvensson:
Closing. Feel free to re-open once the CLA is signed.

patriksvensson:
Superseded by #2162

PR #2162, opened 2026-07-11
Fix infinite loop in TableMeasurer.CollapseWidths when wrappable columns have zero width (#2131)

### LINKED ISSUE #2131: Infinite loop in `TableMeasurer.CollapseWidths` when wrappable columns have width 0

Juimun:
This Pull Request resolves an infinite loop bug within TableMeasurer.CollapseWidths that occurs when one or more wrappable columns already have a width of 0 (e.g., via GetColumnWidth() returning 0 for hidden elements) while non-wrappable columns exceed the maximum console width. This edge case causes the layout engine to freeze, blocking the rendering thread indefinitely.
***

### Fixes #2131

***
 
### Changes
The regression was rooted in the reduction loop calculation when maxColumn dropped to 0. The old logic fallback assigned a value of 1 to secondMaxColumn under certain conditions, yielding a negative columnDifference (0 - 1 = -1). Because the loop termination condition explicitly checked for columnDifference == 0, the negative value bypassed the check, keeping the while loop running forever since Ratio.Reduce cannot perform reductions with negative bounds.

To fully eliminate this lock vector, two changes were introduced to TableMeasurer.cs:
 1. Updated the loop break condition to columnDifference <= 0.
 2. Changed the conditional fallback value inside the secondMaxColumn selector from 1 to 0.
 ***

### DIFF
--- a/src/Spectre.Console.Tests/Unit/Widgets/Table/TableTests.cs
+++ b/src/Spectre.Console.Tests/Unit/Widgets/Table/TableTests.cs
@@ -1056,21 +1056,27 @@ public sealed class TableTests
 
     [Fact(Timeout = 3000)]
     [GitHubIssue("https://github.com/spectreconsole/spectre.console/issues/2131")]
-    public async Task CollapseWidths_Should_Terminate_When_Wrappable_Columns_Collapse_To_Zero()
+    public Task CollapseWidths_Should_Terminate_When_Wrappable_Columns_Collapse_To_Zero()
     {
-        // Given
-        var widths = new List<int> { 0, 0, 100 };
-        var wrappable = new List<bool> { true, true, false };
+        return Task.Run(() =>
+        {
+            // Given
+            var console = new TestConsole().Width(20);
+            var table = new Table().NoBorder();
 
-        var method = typeof(TableMeasurer).GetMethod(
-            "CollapseWidths",
-            System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Static);
+            var col1 = new TableColumn(string.Empty) { Padding = new Padding(0, 0, 0, 0) };
+            var col2 = new TableColumn(string.Empty) { Padding = new Padding(0, 0, 0, 0) };
+            var col3 = new TableColumn(string.Empty) { NoWrap = true, Width = 100, Padding = new Padding(0, 0, 0, 0) };
 
-        // When
-        var result = await Task.Run(() =>
-            (List<int>)method!.Invoke(null, new object[] { widths, wrappable, 5 })!);
+            table.AddColumn(col1);
+            table.AddColumn(col2);
+            table.AddColumn(col3);
 
-        // Then
-        result.ShouldAllBe(w => w >= 0);
+            // When
+            console.Write(table);
+
+            // Then
+            console.Output.ShouldNotBeNullOrEmpty();
+        });
     }
 }
--- a/src/Spectre.Console.Tests/Unit/Widgets/Table/TableTests.cs
+++ b/src/Spectre.Console.Tests/Unit/Widgets/Table/TableTests.cs
@@ -1053,4 +1053,24 @@ public sealed class TableTests
         // Then
         return Verifier.Verify(console.Output);
     }
+
+    [Fact(Timeout = 3000)]
+    [GitHubIssue("https://github.com/spectreconsole/spectre.console/issues/2131")]
+    public async Task CollapseWidths_Should_Terminate_When_Wrappable_Columns_Collapse_To_Zero()
+    {
+        // Given
+        var widths = new List<int> { 0, 0, 100 };
+        var wrappable = new List<bool> { true, true, false };
+
+        var method = typeof(TableMeasurer).GetMethod(
+            "CollapseWidths",
+            System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Static);
+
+        // When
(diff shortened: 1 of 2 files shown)


Juimun:

src/Spectre.Console.Tests/Unit/Widgets/Table/TableTests.cs
```diff
@@ -1053,4 +1053,24 @@ public Task Should_Render_Table_With_Wide_Emoji_Correctly()
...
+            "CollapseWidths",
+            System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Static);
+
+        // When
+        var result = await Task.Run(() =>
+            (List<int>)method!.Invoke(null, new object[] { widths, wrappable, 5 })!);
+
+        // Then
+        result.ShouldAllBe(w => w >= 0);
+    }
```
patriksvensson, line 1081: Great that you added a test for this, but we do not test internals. To test this, recreate a normally failing scenario using the TestConsole, and ensure that it renders.

patriksvensson (changes requested):

Juimun:
Thanks for the feedback! Rewrote the test to use TestConsole and the public API instead of reflection.
To reproduce the [0, 0, X] scenario from the issue without touching internals, I used a NoWrap column with an explicit Width - that bypasses the content measurement clamp in MeasureColumn and lets the wrappable columns collapse to 0, which is what actually triggers the infinite loop in CollapseWidths (the negative columnDifference when maxColumn == 0).
Verified it hangs on the unpatched code and passes once the fix (columnDifference <= 0) is applied.

patriksvensson (approved):

PR #2170, opened 2026-07-20
fix: Validate max value in ProgressTask to ensure it is zero or greater

### LINKED ISSUE #2167: ProgressTask accepts negative MaxValue without validation
### Information

- OS: Ubuntu 24.04
- Version: main (current branch)
- Terminal: GNOME Terminal

### Describe the issue

While exploring the progress API, I noticed that `AddTask` accepts a negative `maxValue` without any validation.

For example:

```csharp
using Spectre.Console;

AnsiConsole.Progress()
    .Start(ctx =>
    {
        var task = ctx.AddTask("Loading World Map", true, -1);

        while (!ctx.IsFinished)
        {
            task.Increment(1);
            Thread.Sleep(50);
        }
    });

The task completes immediately and the progress output appears inconsistent.

While looking through the implementation, I noticed that GetPercentage() explicitly handles MaxValue == 0, but there doesn't seem to be any validation preventing negative values from being passed.

I'm not sure whether accepting negative values is intentional or if they should be rejected (for example by throwing ArgumentOutOfRangeException).

Expected behavior

If negative MaxValue values are not supported, it might be better to validate the argument early instead of allowing the task to enter an inconsistent state.

If this is considered a bug and the proposed direction makes sense, I'd be happy to work on a PR.

Rain6435:
Fixes #2167 

## Changes
This change puts maxValue validation in ProgressTask at the exact points where invalid values can enter task state, so the invariant is enforced centrally instead of at call sites.

Constructor guard in ProgressTask.cs:142: validates initial maxValue when a task is created. This covers all AddTask overloads because they all flow into ProgressTask construction through ProgressContext internals.
Update-path guard in ProgressTask.cs:231: validates later changes via task.MaxValue setter, which routes through Update(maxValue: ...). This prevents a task from becoming invalid after successful creation.
Regression tests in ProgressTests.cs:142 and ProgressTests.cs:167: one test verifies creation-time rejection (ctx.AddTask(..., maxValue: -1)), and one verifies mutation-time rejection (task.MaxValue = -1), so both entry points are locked down.
Putting validation in these two locations ensures consistent runtime behavior for every API path without duplicating checks across ProgressContext overloads or user call sites.

---

### DIFF
--- a/src/Spectre.Console.Tests/Unit/Live/Progress/ProgressTests.cs
+++ b/src/Spectre.Console.Tests/Unit/Live/Progress/ProgressTests.cs
@@ -139,6 +139,57 @@ public sealed class ProgressTests
     }
 
     [Fact]
+    public void Adding_Task_With_Negative_Max_Value_Should_Throw()
+    {
+        // Given
+        var console = new TestConsole()
+            .Interactive();
+
+        var progress = new Progress(console)
+            .Columns(new ProgressBarColumn())
+            .AutoRefresh(false)
+            .AutoClear(false);
+
+        // When
+        var exception = Record.Exception(() =>
+        {
+            progress.Start(ctx =>
+            {
+                ctx.AddTask("foo", maxValue: -1);
+            });
+        });
+
+        // Then
+        exception.ShouldBeOfType<ArgumentOutOfRangeException>();
+    }
+
+    [Fact]
+    public void Setting_Max_Value_To_Negative_Value_Should_Throw()
+    {
+        // Given
+        var console = new TestConsole()
+            .Interactive();
+
+        var progress = new Progress(console)
+            .Columns(new ProgressBarColumn())
+            .AutoRefresh(false)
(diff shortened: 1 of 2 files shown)


Rain6435:

patriksvensson:
Thanks for your contribution!

A question: Wouldn't having a max value of -1 and a min value of -10 be a valid configuration?

Rain6435:
@patriksvensson Good question. In ProgressTask we don’t have a MinValue concept, only Value and MaxValue, so a configuration like max=-1/min=-10 is outside the current model. Indeterminate progress is represented explicitly via IsIndeterminate, and completion is Value >= MaxValue, which means max=-1 would mark a new task as finished immediately. That’s why this change enforces MaxValue >= 0 and keeps indeterminate behavior on the dedicated IsIndeterminate path. Also, this is intuitive way of usage of the ProgressTask.

PR #2178, opened 2026-07-28
fix(TextPrompt): prevent ArgumentOutOfRangeException for non-ASCII default values

### LINKED ISSUE #2152: Status text wraps with incorrect alignment when text exceeds console width
## Description

When using `AnsiConsole.Status()` with text that exceeds the console width, the wrapped continuation lines are not aligned correctly. Instead of wrapping beneath the text column (after the spinner), the text wraps to column 0, causing the spinner and text to overlap visually.

## Expected Behavior

When the status text wraps, continuation lines should be left-aligned with the start of the text (i.e., indented past the spinner column), similar to how other Spectre.Console widgets handle multi-line content.

## Actual Behavior

The wrapped text starts at column 0 on the next line, resulting in misaligned output. The spinner character appears to overlap with or push into the wrapped text.

## Steps to Reproduce

```csharp
using Spectre.Console;

// Use a narrow console or a long message that exceeds the terminal width
var longMessage = "Building AppHost... playground\\HealthChecks\\HealthChecksSandbox.AppHost\\AppHost.csproj";

await AnsiConsole.Status()
    .Spinner(Spinner.Known.Dots3)
    .StartAsync(longMessage, async ctx =>
    {
        await Task.Delay(5000);
    });
```

Run this in a terminal narrower than the message length (e.g., 80 columns), or use a very long path to trigger wrapping.

## Recording

https://github.com/user-attachments/assets/74e1f5ec-b542-44f4-8ed6-ee5b569d8314

Displays:
1. Status where text doesn't wrap. Looks good
2. Status where text does wrap. Unexpected display of text. I'd expect it to wrap underneath the existing text to the left of the status indicator.
    Image

## Environment

- Spectre.Console version: 0.50.0
- OS: Windows 11
- Terminal: Windows Terminal

Moha-sami:
Fixes #2152

### Problem
When using `TextPrompt` with `EditableDefaultValue(true)` containing non-ASCII or Unicode characters (e.g. `"日本語"` or `"ä"`), `AnsiConsoleExtensions.Input.cs` passed `(ConsoleKey)ch` to `ConsoleKeyInfo`, which throws `ArgumentOutOfRangeException` in .NET when `ConsoleKey > 255`.

### Solution
1. Updated `AnsiConsoleExtensions.Input.cs` to check `(int)ch <= 255 ? (ConsoleKey)ch : (ConsoleKey)0`, preventing out-of-range exceptions while maintaining full Unicode string processing.
2. Added unit test `Should_Not_Throw_When_Editable_Default_Value_Contains_Non_Ascii_Characters` in `TextPromptTests.cs`.

### DIFF
--- a/src/Spectre.Console.Tests/Unit/Prompts/TextPromptTests.cs
+++ b/src/Spectre.Console.Tests/Unit/Prompts/TextPromptTests.cs
@@ -53,6 +53,23 @@ public sealed class TextPromptTests
     }
 
     [Fact]
+    public void Should_Not_Throw_When_Editable_Default_Value_Contains_Non_Ascii_Characters()
+    {
+        // Given
+        var console = new TestConsole();
+        console.Input.PushKey(ConsoleKey.Enter);
+
+        // When
+        var result = console.Prompt(
+            new TextPrompt<string>("Language?")
+                .DefaultValue("日本語")
+                .EditableDefaultValue(true));
+
+        // Then
+        result.ShouldBe("日本語");
+    }
+
+    [Fact]
     [Expectation("InvalidChoice")]
     public Task Should_Return_Error_If_An_Invalid_Choice_Is_Made()
     {
--- a/src/Spectre.Console/Extensions/AnsiConsoleExtensions.Input.cs
+++ b/src/Spectre.Console/Extensions/AnsiConsoleExtensions.Input.cs
@@ -22,7 +22,8 @@ public static partial class AnsiConsoleExtensions
             foreach (var ch in initialInput)
             {
                 var control = char.IsUpper(ch);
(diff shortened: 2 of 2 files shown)


patriksvensson:
@Moha-sami That issue you mention had been closed, and is not related to this change.

patriksvensson:
This PR needs a related issue.

Also, are you sure no AI has been used in this change? We need to know if so.

Moha-sami:
No But I'm new to This some how It's ok Will Just remove it my contribution was nothing anyway Good Luck

patriksvensson:
It's OK being new, but in the future, try following the contribution guidelines for projects that you want to contribute to.

PR #2181, opened 2026-07-29
Support disabled items in SelectionPrompt<T>

### LINKED ISSUE #2180: Support disabled/unselectable items in SelectionPrompt<T>
See #2179 for background.

I'm working on a project where a `SelectionPrompt<T>` needs to show choices that are temporarily unavailable (e.g. an action that can't currently be used), visible in the list for context, but not navigable to or selectable. Right now there's no way to do this: `ISelectionItem<T>` has no per-item enabled/disabled state, and `DisabledStyle` is only ever applied to group headers in `Mode.Leaf`.

Proposed API: a separate opt-in interface rather than adding members to `ISelectionItem<T>` directly, so it's additive on every target framework, including `netstandard2.0`, which can't represent default interface implementations at the runtime level:

```csharp
public interface IDisableableSelectionItem<T> : ISelectionItem<T>
    where T : notnull
{
    bool IsDisabled { get; }

    IDisableableSelectionItem<T> Disable();
}
```

`ListPromptItem<T>` implements it alongside `IMultiSelectionItem<T>`. An extension method keeps the call site unchanged:

```csharp
public static ISelectionItem<T> Disable<T>(this ISelectionItem<T> item)
    where T : notnull
{
    if (item is IDisableableSelectionItem<T> disableable)
    {
        disableable.Disable();
    }

    return item;
}
```

```csharp
prompt.AddChoice("Fireball").Disable();
```

`ISelectionItem<T>` itself is untouched, so this doesn't affect any existing implementation of it, on any target framework.

I've prototyped this locally. It extends `ListPromptItem<T>`, `SelectionPrompt<T>`'s render/input handling, and `ListPromptState<T>`'s navigation/search filtering to treat disabled leaves the same way group headers already are (styled, skipped during arrow-key nav and search, blocked on submit). Small, additive diff, verified building and passing tests across all four target frameworks (net8.0/net9.0/net10.0/netstandard2.0), and I've added tests covering the navigation/search/default-value edge cases. Ready to open as a PR against this issue.

thomas-mathers:
Closes #2180.

Adds `IDisableableSelectionItem<T>`, a separate opt-in interface rather than new members on `ISelectionItem<T>` directly, so it stays additive on every target framework (including `netstandard2.0`, which can't represent default interface implementations). `ListPromptItem<T>` implements it alongside `IMultiSelectionItem<T>`, and a `SelectionItemExtensions.Disable()` extension keeps the call site unchanged:

```csharp
prompt.AddChoice("Fireball").Disable();
```

A disabled item is styled with `DisabledStyle` and skipped during arrow-key navigation, search, and default-value selection, the same way group headers already are in `Mode.Leaf`.

Opening as a draft since I haven't heard back on the issue write-up yet, happy to take feedback on the API shape before marking ready.

## Test plan
- [x] Builds clean across all four target frameworks (net8.0/net9.0/net10.0/netstandard2.0)
- [x] Existing test suite passes (756/756)
- [x] New tests added covering navigation skip, wrap-around skip, search exclusion, and default-value fallback for disabled items

### DIFF
--- a/src/Spectre.Console.Tests/Expectations/Public_API.Output.verified.txt
+++ b/src/Spectre.Console.Tests/Expectations/Public_API.Output.verified.txt
@@ -2242,6 +2242,12 @@
         bool NoWrap { get; set; }
         int? Width { get; set; }
     }
+    public interface IDisableableSelectionItem<T> : Spectre.Console.ISelectionItem<T>
+        where T :  notnull
+    {
+        bool IsDisabled { get; }
+        Spectre.Console.IDisableableSelectionItem<T> Disable();
+    }
     public interface IExclusivityMode
     {
         T Run<T>(System.Func<T> func);
@@ -2769,6 +2775,11 @@
     {
         public static Spectre.Console.Rule RuleStyle(this Spectre.Console.Rule rule, Spectre.Console.Style style) { }
         public static Spectre.Console.Rule RuleTitle(this Spectre.Console.Rule rule, string title) { }
+    }
+    public static class SelectionItemExtensions
+    {
+        public static Spectre.Console.ISelectionItem<T> Disable<T>(this Spectre.Console.ISelectionItem<T> item)
+            where T :  notnull { }
     }
     public enum SelectionMode
     {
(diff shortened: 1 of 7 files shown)


thomas-mathers:

patriksvensson:
Any specific reason you closed this?

PR #2183, opened 2026-07-30
fix: IsFinished should return false when a task has not started

### LINKED ISSUE #2182: ProgressContext.IsFinished returns true when tasks exist but none have started
**Information**
- OS: Ubuntu 24.04
- Version: main (current branch)
- Terminal: GNOME Terminal

**Describe the bug**
`ProgressContext.IsFinished` returns `true` when one or more tasks exist in the context but none of them have started yet (for example, when created with `autoStart: false`). This causes any `while (!ctx.IsFinished)` loop to be skipped entirely, even though the intended work was never started.

**To Reproduce**

```csharp
using Spectre.Console;

AnsiConsole.Progress().Start(ctx =>
{
    var task = ctx.AddTask("Loading", autoStart: false);

    while (!ctx.IsFinished)
    {
        Thread.Sleep(50);
        task.StartTask();
        Console.WriteLine("Im here");
    }
});
```

1. Run the code above.
2. Observe that `Im here` is never printed.
3. The process exits with code 0, and the progress bar stays frozen at 0%.

Looking at the implementation of `IsFinished`:

```csharp
public bool IsFinished
{
    get
    {
        lock (_taskLock)
        {
            return _tasks.Where(x => x.IsStarted).All(task => task.IsFinished);
        }
    }
}
```

When no task has `IsStarted == true`, `Where(...)` produces an empty sequence, and `.All()` on an empty sequence always returns `true` (vacuous truth in LINQ). So `IsFinished` reports "finished" even though nothing has actually run yet.

**Expected behavior**
There are two reasonable directions here, and I'd like maintainer input on which is intended:

1. `IsFinished` should return `false` when the context has tasks that haven't started yet, since there's still pending work — only a truly empty context (`_tasks.Count == 0`) would return `true`.
2. The current behavior is intentional, and the XML doc should be updated to make this edge case explicit (currently: "Gets a value indicating whether or not all started tasks have completed", which doesn't call out the "no tasks started" case).

I'd lean towards option 1, since it matches the common `while (!ctx.IsFinished)` usage pattern. If this is confirmed as a bug, I'd be happy to work on a PR with regression tests covering both the "no tasks added" and "tasks added but not started" cases.

MateusMo:
Fixes #2182

- [x] I have read the Contribution Guidelines

## Changes

`ProgressContext.IsFinished` previously used `_tasks.Where(x => x.IsStarted).All(task => task.IsFinished)`. When no task in the context had started yet (for example, tasks created with `autoStart: false`), `Where(...)` produced an empty sequence, and `.All()` on an empty sequence always returns `true` (vacuous truth). This caused `IsFinished` to report `true` even though no work had actually run, silently breaking the common `while (!ctx.IsFinished)` usage pattern.

- `ProgressContext.cs`: `IsFinished` now returns `true` only when there are no tasks in the context, or when every task's own `IsFinished` is `true`. Since `ProgressTask.IsFinished` is `false` for any task that hasn't reached its `MaxValue` (including one that hasn't started), this naturally treats "not started" as "not finished" without needing a separate `IsStarted` check.
- `ProgressTests.cs`: added four regression tests covering: no tasks in the context, a single task with `autoStart: false`, a task that has started and finished, and a mix of one finished task alongside one not-started task.

---

### DIFF
--- a/src/Spectre.Console.Tests/Unit/Live/Progress/ProgressTests.cs
+++ b/src/Spectre.Console.Tests/Unit/Live/Progress/ProgressTests.cs
@@ -727,4 +727,100 @@ public sealed class ProgressTests
                 "          \n" + // Bottom padding
                 "\e[?25h"); // show cursor
     }
+
+    [Fact]
+    public void IsFinished_Should_Be_True_When_No_Tasks_Exist()
+    {
+        // Given
+        var console = new TestConsole().Interactive();
+        var progress = new Progress(console)
+            .Columns(new ProgressBarColumn())
+            .AutoRefresh(false)
+            .AutoClear(false);
+
+        var result = false;
+
+        // When
+        progress.Start(ctx =>
+        {
+            result = ctx.IsFinished;
+        });
+
+        // Then
+        result.ShouldBeTrue();
+    }
+
+    [Fact]
+    public void IsFinished_Should_Be_False_When_Task_Has_Not_Started()
+    {
+        // Given
+        var console = new TestConsole().Interactive();
+        var progress = new Progress(console)
+            .Columns(new ProgressBarColumn())
+            .AutoRefresh(false)
+            .AutoClear(false);
+
+        var result = false;
+
+        // When
+        progress.Start(ctx =>
+        {
+            ctx.AddTask("foo", autoStart: false);
+            result = ctx.IsFinished;
+        });
+
+        // Then
+        result.ShouldBeFalse();
+    }
+
+    [Fact]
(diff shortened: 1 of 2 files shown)


MateusMo:

patriksvensson:
Thanks!

Due to the very long "Changes" section I need to ask; there was no AI involved in this PR?

MateusMo:
Yes, I found the behavior running a console app calling the method and used claude ai to help me to validate the scenarios and improve the test coverage

MateusMo:
Any update on this one? Happy to make changes if needed

PR #2196, opened 2026-09-01
Add IncrementMaxValue method to ProgressTask (fixes #1893)

### LINKED ISSUE #1893: Add IncrementMaxValue method to ProgressTask for thread safety
A feature.  

Updating the `MaxValue` property is not thread safe, and no thread safe method is available to do this. This is a common scenario where I might be 'discovering' any number of items that the task then has to perform some operation on.

This will lose progress updates:
```
var progress = new Progress<DiscoveryProgress>(count =>
{
    someTask.MaxValue += count;
}
await someParallelMethod(progress);
```

A workaround for this is:
```
var maxValue = 0;
var progress = new Progress<DiscoveryProgress>(count =>
{
    someTask.MaxValue = Interlocked.Add(count);
}
await someParallelMethod(progress);
```

Which is functional, but can become a problem with many tasks.

However, thread safety is already in place for updating the `Value` property via `Increment`.

```
    public void Increment(double value)
    {
        Update(increment: value);
    }

    private void Update(
        string? description = null,
        double? maxValue = null,
        double? increment = null,
        double? value = null)
    {
        lock (_lock)
        {
            // Omitted code here
        }
    }
```

`Update` is already thread safe, and even has the capability to update the `MaxValue` property, but is inaccessible due to it being private.

An additional public method in `ProgressTask`: `IncrementMaxValue` that uses the existing private `Update` method to allow for thread safe updates to `MaxValue`.

Workaround described above.

**Additional context**
none

---

Margalit111:
Fixes #1893

Assisted by AI (Gemini) for implementation guidance and unit test drafting.

## Changes

- Added `IncrementMaxValue` method to `ProgressTask` to allow thread-safe updates to the task's maximum value.
- Updated the internal `Update` method to safely modify `_maxValue` within the lock.
- Added fluent extension method `IncrementMaxValue` to `ProgressTaskExtensions`.
- Added unit tests in `ProgressTests.cs` verifying `IncrementMaxValue` behavior.

### DIFF
--- a/src/Spectre.Console.Tests/Expectations/Public_API.Output.verified.txt
+++ b/src/Spectre.Console.Tests/Expectations/Public_API.Output.verified.txt
@@ -2681,6 +2681,7 @@
         public object? Tag { get; set; }
         public double Value { get; set; }
         public void Increment(double value) { }
+        public void IncrementMaxValue(double value) { }
         public void StartTask() { }
         public void StopTask() { }
     }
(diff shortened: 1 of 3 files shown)


Margalit111:

